PR #8, opened 2016-12-14
Add an explanatory comment to MersenneTwisterEngine.opCall

WebDrake:
This comment covers the essentials of how `opCall` is structured: the two separate procedures that are carried out, and the importance of them being interwoven rather than performed sequentially.

@9il hope this is useful -- I felt that your explanation of the design choices made here shouldn't be limited to a forum post.


9il:
Thanks!

PR #12, opened 2016-12-28
[WIP][DO NOT MERGE] Test OS random APIs

wilzbach:
Hi,

so this is WIP and while it's pretty easy to read random number in Posix, the Windows part is troubling me. While [there's an API](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379942(v=vs.85).aspx) which many implementation use:

C++: http://en.cppreference.com/w/cpp/numeric/random/random_device
LLVM: https://github.com/llvm-mirror/libcxx/blob/master/src/random.cpp
G++: https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/src/c%2B%2B11/random.cc
Python urandom: https://github.com/python/cpython/blob/master/Python/random.c
Vibe.d cryptorand: https://github.com/rejectedsoftware/vibe.d/blob/master/crypto/vibe/crypto/cryptorand.d
pycrypto: https://github.com/dlitz/pycrypto/blob/master/src/winrand.c

I can't get it working on the AppVeyor test instance. For experimentation I also copied `cryptorand` from Vibe.d over and it's unittests are failing as well. I always get `ERROR_INVALID_PARAMETER` which according to the docs stands for:

> One of the parameters contains a value that is not valid. This is most often a pointer that is not valid.

To be honest I haven't used a Windows machine for ages, so if someone has more experience with Windows - it would be greatly appreciated.
Could it be that the linking with `pragma(lib, "Advapi32.lib")` does sth. weird?
Unfortunately specifying the libs via dub didn't work, e.g.:

```
"libs-windows-x86-dmd": ["Advapi32.lib"],
"libs-windows-x86_64-dmd": ["Advapi32.lib"]
```

### DIFF
--- a/source/mir/random/engine/test.d
+++ b/source/mir/random/engine/test.d
@@ -128,16 +128,26 @@ version(Posix)
 
 version(Windows)
 {
+    // the windows header in core.runtime are broken for x64!
+    alias ULONG_PTR = size_t;
+    alias BOOL = bool;
+    alias DWORD = uint;
+    alias LPCWSTR = wchar*;
+    alias PBYTE = ubyte*;
+    alias HCRYPTPROV = ULONG_PTR;
+    alias LPCSTR = const(char)*;
 
-    version(DigitalMars)
-    pragma(lib, "Advapi32.lib");
+    import core.sys.windows.wincrypt : PROV_RSA_FULL, CRYPT_NEWKEYSET, CRYPT_VERIFYCONTEXT, CRYPT_SILENT;
 
-    import core.sys.windows.wincrypt : HCRYPTPROV;
-    private static HCRYPTPROV hProvider;
+    extern(Windows) BOOL CryptGenRandom(HCRYPTPROV, DWORD, PBYTE);
+    extern(Windows) BOOL CryptAcquireContextA(HCRYPTPROV*, LPCSTR, LPCSTR, DWORD, DWORD);
+    extern(Windows) BOOL CryptAcquireContextW(HCRYPTPROV*, LPCWSTR, LPCWSTR, DWORD, DWORD);
+    extern(Windows) BOOL CryptReleaseContext(HCRYPTPROV, ULONG_PTR);
+
+    private static ULONG_PTR hProvider;
 
     auto initGetRandom()
     {
-        import core.sys.windows.wincrypt;
         import core.sys.windows.winbase : GetLastError;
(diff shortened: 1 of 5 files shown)


source/mir/random/engine/test.d
```diff
@@ -0,0 +1,232 @@
...
+     * Flags for getrandom(2)
+     *
+     * GRND_NONBLOCK	Don't block and return EAGAIN instead
+     * GRND_RANDOM		Use the /dev/random pool instead of /dev/urandom
+     */
+    enum GRND_NONBLOCK = 0x0001;
+    enum GRND_RANDOM = 0x0002;
+    enum GETRANDOM = 318;
+
+    size_t syscall(size_t ident, size_t n, size_t arg1, size_t arg2)
```
9il: Why not to use C `getrandom` instead?

source/mir/random/engine/test.d
```diff
@@ -0,0 +1,232 @@
...
+    private static bool hasGetRandom;
+
+    auto initHasGetRandom()
+    {
+        import core.sys.posix.sys.utsname;
+        utsname uts;
+        uname(&uts);
+
+        import std.algorithm.iteration : splitter;
+        import std.conv : to;
```
9il: Please avoid dependencies from std.algorithm and std.conv
C functions can be used instead

wilzbach: I know - this is only the prototype ;-)

source/mir/random/engine/test.d
wilzbach: Because it's only a syscall:

http://man7.org/linux/man-pages/man2/getrandom.2.html

wilzbach:
(closing temporarily)

source/mir/random/engine/test.d
9il: Do you mean we can not use C `getrandom` instead>

9il: `syscall` implementation is platform dependent and requires DRuntime because `synchronization`

wilzbach: > Do you mean we can not use C getrandom instead?

It's not part of the C library - it's only available as `syscall`.

> syscall implementation is platform dependent

It's only needed/used for `Linux x86_64`.

> syscall implementation is requires DRuntime because synchronization

Oh - can we do it without `synchronized`?

PR #13, opened 2016-12-28
Get random data from OS APIs

wilzbach:
So I managed to find the issue in the Windows API. It was due to wrong API headers in druntime - for 64-bit pointers need to have 8 bytes.

Let me summarize:

Current Behavior
----------------------

- Windows -> [`CryptGenRandom`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379942(v=vs.85).aspx)
- Linux
   - `>= 3.17` -> [`getrandom`](http://man7.org/linux/man-pages/man2/getrandom.2.html) system call, [intro article](https://lwn.net/Articles/605828)
   - otherwise: `/dev/(u)?random` (see e.g. [`/dev/random` vs. `/dev/urandom`](http://www.2uo.de/myths-about-urandom))
- MacOS/OpenBSD/Solaris: `/dev/(u)?random` (see e.g. [`/dev/random` vs. `/dev/urandom`](http://www.2uo.de/myths-about-urandom))

Todo
-------

- [x] decide on an API
- [x] write more documentation
- [x] add support for D attributes
- [x] add tests

Future work
---------------

- OpenBSD -> [`getentropy`](http://man.openbsd.org/getentropy.2) (Not implemented as I don't have a OpenBSD atm - anyone?)

References
---------------

C++ API: http://en.cppreference.com/w/cpp/numeric/random/random_device
LLVM stdcxx: https://github.com/llvm-mirror/libcxx/blob/master/src/random.cpp
G++: https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/src/c%2B%2B11/random.cc
Python `urandom`: https://github.com/python/cpython/blob/master/Python/random.c
Vibe.d `cryptorand`: https://github.com/rejectedsoftware/vibe.d/blob/master/crypto/vibe/crypto/cryptorand.d
`pycrypto`: https://github.com/dlitz/pycrypto/blob/master/src/winrand.c


source/mir/random/engine/test.d
```diff
@@ -0,0 +1,236 @@
...
+
+        asm
+        {
+            mov RAX, ident;
+            mov RDI, n[RBP];
+            mov RSI, arg1[RBP];
+            mov RDX, arg2[RBP];
+            syscall;
+            mov ret, RAX;
+        }
```
wilzbach: @9il Is `dmd` guaranteed to execute the `asm` block in one go?

9il: What do you mean?

wilzbach: I got the inspiration to use this `syscall` style from here: https://github.com/kubo39/syscall-d/blob/master/source/syscalld/arch/syscall_x86_64.d
Hence also the `synchronized` statements, but afaict we don't need it - e.g.:
https://wiki.dlang.org/LDC_inline_assembly_expressions#X86-64

wilzbach:
I accidentally came by two other implementation in the D ecosystem:

https://github.com/LightBender/SecureD/blob/master/source/random.d
https://github.com/etcimon/botan/blob/master/source/botan/libstate/libstate.d#L190

(the latter is interesting as it uses quite a bunch of entropy sources)

9il:
> I accidentally came by two other implementation in the D ecosystem:

Awesome!

9il:
But do not spent much time on deprecated targets ;)

9il:
@wilzbach  do you need help with this PR?

wilzbach:
> @wilzbach do you need help with this PR?

Oh sorry for letting this get stalled.

As discussed this exposes two different functions with C-style like interface, but with D-guarantees in terms of `@safe nothrow @nogc`: 

- `getRandomBlocking`
- `getRandomNonBlocking`

The difference between these two functions is that the `Blocking` one will `loop` until the buffer is filled (i.e. until enough entropy is generated), while the NonBlocking one will yield an error code.
If initialization needs to be done, it's done on the first call.

@9il please let me know if there's sth. missing for this PR.

About Blocking behavior:
----------------------------------

tl;dr: only useful on Linux.

- Linux docs state: 

> The /dev/random device is a legacy interface which dates back to a
       time where the cryptographic primitives used in the implementation of
       /dev/urandom were not widely trusted.  It will return random bytes
       only within the estimated number of bits of fresh noise in the
       entropy pool, blocking if necessary.  /dev/random is suitable for
       applications that need high quality randomness, and can afford
       indeterminate delays.

> When the entropy pool is empty, reads from /dev/random will block
       until additional environmental noise is gathered.

And for the newer `getRandom`:

>     GRND_NONBLOCK
              By default, when reading from the random source, getrandom()
              blocks if no random bytes are available, and when reading from
              the urandom source, it blocks if the entropy pool has not yet
              been initialized.  If the GRND_NONBLOCK flag is set, then
              getrandom() does not block in these cases, but instead
              immediately returns -1 with errno set to EAGAIN.

- Windows doesn't support _blocking_ behavior (the documentation doesn't even state how `CryptGenRandom` behaves if the entropy pool is empty)
- macOS doesn't [support _blocking_ either](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man4/random.4.html) (according to the [docs](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man4/random.4.html) it isn't necessary to block as a special `SecurityServer` takes care of ensuring that there's always enough entropy)

source/mir/random/engine/package.d
```diff
@@ -68,60 +68,10 @@ A single unsigned integer seed value, different on each successive call
...
-    }
-    ulong k = ((cast(ulong)pid << 32) ^ tid) + ticks;
-    k ^= k >> 33;
-    k *= 0xff51afd7ed558ccd;
-    k ^= k >> 33;
-    k *= 0xc4ceb9fe1a85ec53;
-    k ^= k >> 33;
-    return cast(size_t)k;
+    size_t seed;
+    //getRandomBlocking(&seed, seed.sizeof);
```
9il: 1. Lets preserve old version if `getRandomBlocking` returns -1
2. and uncomment this line)

source/mir/random/engine/package.d
```diff
@@ -151,3 +101,367 @@ unittest
...
+        AVAILABLE,
+    }
+
+    // getrandom was introduced in 3.17
+    private static GET_RANDOM hasGetRandom = GET_RANDOM.UNINTIALIZED;
+
+    import core.sys.posix.sys.utsname : utsname;
+
+    // druntime isn't properly annotated
+    extern(C) int uname(utsname* __name) @nogc nothrow;
```
9il: private

source/mir/random/engine/package.d
```diff
@@ -151,3 +101,367 @@ unittest
...
+        {
+            token = strtok(p, ".");
+            if (atoi(token) >= 17)
+                return true;
+        }
+
+        return false;
+    }
+
+    extern(C) int syscall(size_t ident, size_t n, size_t arg1, size_t arg2) @nogc nothrow;
```
9il: private

source/mir/random/engine/package.d
```diff
@@ -151,3 +101,367 @@ unittest
...
+        sizes.
+    */
+    private ptrdiff_t getRandomImplSysBlocking(void* ptr, size_t len) @nogc @trusted nothrow
+    {
+        auto genBytes = 0;
+        while (genBytes < len)
+        {
+            auto res = syscall(GETRANDOM, cast(size_t) ptr + genBytes, len - genBytes, 0);
+            if (res >= 0)
+                genBytes -= res;
```
9il: bug.
Just move ptr and reduce length, instead of `genBytes` variable

source/mir/random/engine/package.d
```diff
@@ -151,3 +101,367 @@ unittest
...
+        return syscall(GETRANDOM, cast(size_t) ptr, len, GRND_NONBLOCK);
+    }
+}
+
+version(Posix)
+{
+    import core.stdc.stdio : fclose, feof, ferror, fopen, fread;
+    alias IOType = typeof(fopen("a", "b"));
+    static private IOType fdRandom;
+    static private IOType fdURandom;
```
9il: `__gshared`

9il: `private`

source/mir/random/engine/package.d
```diff
@@ -151,3 +101,367 @@ unittest
...
+
+    shared static ~this()
```
9il: `extern(C)`

source/mir/random/engine/package.d
```diff
@@ -151,3 +101,367 @@ unittest
...
+    private alias DWORD = size_t; // uint in druntime
+    private alias LPCWSTR = wchar*;
+    private alias PBYTE = ubyte*;
+    private alias HCRYPTPROV = ULONG_PTR;
+    private alias LPCSTR = const(char)*;
+
+    extern(Windows) BOOL CryptGenRandom(HCRYPTPROV, DWORD, PBYTE) @nogc @safe nothrow;
+    extern(Windows) BOOL CryptAcquireContextA(HCRYPTPROV*, LPCSTR, LPCSTR, DWORD, DWORD) @nogc nothrow;
+    extern(Windows) BOOL CryptAcquireContextW(HCRYPTPROV*, LPCWSTR, LPCWSTR, DWORD, DWORD) @nogc nothrow;
+    extern(Windows) BOOL CryptReleaseContext(HCRYPTPROV, ULONG_PTR) @nogc nothrow;
```
9il: private

source/mir/random/engine/package.d
```diff
@@ -151,3 +101,367 @@ unittest
...
+            else
+            {
+                return 1;
+            }
+        }
+
+        return 0;
+    }
+
+    shared static ~this()
```
9il: extern(C)

source/mir/random/engine/package.d
```diff
@@ -151,3 +101,367 @@ unittest
...
+    private alias PBYTE = ubyte*;
+    private alias HCRYPTPROV = ULONG_PTR;
+    private alias LPCSTR = const(char)*;
+
+    extern(Windows) BOOL CryptGenRandom(HCRYPTPROV, DWORD, PBYTE) @nogc @safe nothrow;
+    extern(Windows) BOOL CryptAcquireContextA(HCRYPTPROV*, LPCSTR, LPCSTR, DWORD, DWORD) @nogc nothrow;
+    extern(Windows) BOOL CryptAcquireContextW(HCRYPTPROV*, LPCWSTR, LPCWSTR, DWORD, DWORD) @nogc nothrow;
+    extern(Windows) BOOL CryptReleaseContext(HCRYPTPROV, ULONG_PTR) @nogc nothrow;
+
+    private static ULONG_PTR hProvider;
```
9il: `__gshared`

wilzbach:
> Each public function should be extern(C) and starts with mir_random_. mir_random_genRandomBlocking

Ok - questions:

- It's currently in the `mir.random.engine` - so shouldn't it be `mir_random_engine_genRandomBlocking`?
- This will look ugly on the documentation pages. Any trick for this? Can't we use `extern(C)` on an `alias`?
- What about the module deconstructor? Should we expose sth. like `mir_random_engine_dtor`?

> `__gshared`

I assume I can't use any locking in betterC?
There are four `__gshared` variables then. In case of a race:

- `hasGetRandom`: Linux kernel version will be calculated twice (will always yield the same) -> not a problem
- `hProvider`: we could acquire two Windows Crypto contexts (leak)
- `fdRandom` / `fdURandom`: we could open the file twice and leak a file handle

We could provide a special `init` method which inits the handles and ensures that no leaks can occur. On the D side it could automatically be called with the shared module constructor

source/mir/random/engine/package.d
```diff
@@ -151,3 +143,378 @@ unittest
...
+*/
+extern(C) ptrdiff_t mir_random_genRandomBlocking(void* ptr , size_t len) @nogc @trusted nothrow
+{
+    version(Windows)
+    {
+        if (hProvider == 0)
+            if (initGetRandom != 0)
+                return -1;
+
+        while(!CryptGenRandom(hProvider, len, cast(PBYTE) ptr)) {}
```
wilzbach: For Windows all [error codes](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379942(v=vs.85).aspx) of `CryptGenRandom` seem unrecoverable, so the loop here doesn't seem very useful. I think we should directly yield the error here.

9il: ok

9il:
> It's currently in the mir.random.engine - so shouldn't it be mir_random_engine_genRandomBlocking?

no

> This will look ugly on the documentation pages. Any trick for this? Can't we use extern(C) on an alias?

This should be a part of documentation.

> What about the module deconstructor? Should we expose sth. like mir_random_engine_dtor?

Yes, but `extern(C)` should be added to the D's module destructor.

9il:
> We could provide a special init method which inits the handles and ensures that no leaks can occur. On the D side it could automatically be called with the shared module constructor

`mir_random_engine_ctor ` and the `extern(C)` D's module constructor

9il:
Hey @wilzbach , could you please finalise this PR?

wilzbach:
> Hey @wilzbach , could you please finalise this PR?

Sorry that I left this pending for such a long time.

> `mir_random_engine_ctor ` and the `extern(C)` D's module constructor

Added :)

9il:
@wilzbach please convert all private functions to templates and add `version(LDC) pragma(inline, true);` for them. This will make object files clearer.

PR #14, opened 2017-01-07
Update MersenneTwisterEngine template signature to match C++11 standard

WebDrake:
The original template signature diverges from the C++11 standard both in names and types of template parameters.  While not producing a functional difference, matching the standard allows for a less ambiguous declaration and makes for easier adaptation on the part of users used to C++ code.

These patches also introduce the previously-missing "initialization multiplier" template parameter `f`.  Since the correct value for this parameter is not solely dependent on the unsigned integer type in use, the original `static if` would not have covered all possible instantiations of the generator.

Finally, a basic unittest has been added for `Mt19937_64` both to validate the generator and ensure that its behaviour does not change as a result of these updates.

### DIFF
--- a/source/mir/random/engine/mersenne_twister.d
+++ b/source/mir/random/engine/mersenne_twister.d
@@ -15,7 +15,7 @@ The $(LUCKY Mersenne Twister) generator.
 struct MersenneTwisterEngine(UIntType, size_t w, size_t n, size_t m, size_t r,
                              UIntType a, size_t u, UIntType d, size_t s,
                              UIntType b, size_t t,
-                             UIntType c, size_t l)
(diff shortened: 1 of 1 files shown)


WebDrake:
@9il I'm submitting this PR among other things because I think these changes will create an easier basis from which to address the issues raised in https://github.com/dlang/phobos/pull/5011#issuecomment-270804028 (it means we can more easily make a direct comparison of behaviour to C++11 implementations).

9il:
Thanks @WebDrake !

WebDrake:
You're welcome.  Give me a few minutes and I'll write up an issue on the seeding/word size issue mentioned in the Phobos PR.

Issue #15, opened 2017-01-07
MersenneTwisterEngine does not support non-standard word sizes


WebDrake:
The `MersenneTwisterEngine` template distinguishes between the `UIntType` used to store generator state, and the word size `w`: for example, when `UIntType` is `ulong` (64-bit), the word size may still be only 32 or (say) 48 bits.  In this case the returned variates should consist of only the lowest `w` bits of the generated `UIntType` value (which is not currently enforced) and the seeding process should also take this into account (currently only partially done: the first entry in the state array takes the lowest `w` bits of the provided seed, but the population of the other state array entries does not take this into account).

Comparison of output of `mir.random`'s `MersenneTwisterEngine` and C++11's `mersenne_twister_engine` can be made using the following D unittest:

```D
unittest
{
    alias MT(UIntType, uint w) = MersenneTwisterEngine!(UIntType, w, 624, 397, 31,
                                                        0x9908b0df, 11, 0xffffffff, 7,
                                                        0x9d2c5680, 15,
                                                        0xefc60000, 18, 1812433253);

    foreach (R; std.meta.AliasSeq!(MT!(uint, 32), MT!(ulong, 32), MT!(ulong, 48), MT!(ulong, 64)))
    {
        static if (R.wordSize == 48) static assert(R.max == 0xFFFFFFFFFFFF);
        auto a = R(R.defaultSeed);
        import std.stdio : writeln;
        writeln(a());
        writeln(a());
        writeln();
    }
}
```

... and the following C++11 program:

```C++
#include <cinttypes>
#include <iostream>
#include <random>
#include <type_traits>

template<class UIntType, size_t w>
void mt_result ()
{
    std::mersenne_twister_engine<
        UIntType, w, 624, 397, 31,
        0x9908b0df, 11, 0xffffffff, 7,
        0x9d2c5680, 15,
        0xefc60000, 18, 1812433253> gen;

    gen.seed(std::mt19937::default_seed);
    std::cout << gen() << std::endl;
    //for (int i = 0; i < 599; ++i)
    //  gen();
    std::cout << gen() << std::endl;
    std::cout << std::endl;
}

int main ()
{
    mt_result<uint32_t, 32>();
    mt_result<uint64_t, 32>();
    mt_result<uint64_t, 48>();
    mt_result<uint64_t, 64>();
}
```

WebDrake:
[...]

PR #16, opened 2017-01-07
fix issue #15

9il:


9il:
ping @WebDrake

WebDrake:
Looks OK in principle.  In some ways I'm not sure I'm comfortable with dual-purposing `max` as both a value and as a bitmask, but your call.

source/mir/random/engine/mersenne_twister.d
```diff
@@ -30,8 +30,8 @@ struct MersenneTwisterEngine(UIntType, size_t w, size_t n, size_t m, size_t r,
     @disable this();
     @disable this(this);
 
-    private enum UIntType upperMask = ~((cast(UIntType) 1u << (UIntType.sizeof * 8 - (w - r))) - 1);
     private enum UIntType lowerMask = (cast(UIntType) 1u << r) - 1;
+    private enum UIntType upperMask = ~lowerMask;
```
WebDrake: Note here: according to https://en.wikipedia.org/wiki/Mersenne_Twister `upperMask` is supposed to be the lowest `w` bits of `~lowerMask`.

9il: Thanks! fixed in #17

WebDrake:
Last comment I'll make on here: isn't the return value of `opCall` also supposed to be the lowest `w` bits of the generated value?  (It might be superfluous if all is set up correctly but my inclination is to be safe with this.)

9il:
Yes, but if `b` and `c` is correct, then `z` is correct too without `& max`

WebDrake:
What are the constraints on `b` and `c` that would allow us to determine that they are correct in this respect?

9il:
They should be less then or equal to `max` (fit into lowest `w` bits)

9il:
the check is already here: `static assert(a <= max && b <= max && c <= max);`

WebDrake:
OK, cool.  Thanks for the quick fixes -- I'll get them over into the phobos PR soon!

One remark -- I didn't include any constraints on `f` in my earlier PR but if I've understood correctly, the same principle applies: it should be less than or equal to `max`.

9il:
> One remark -- I didn't include any constraints on f in my earlier PR but if I've understood correctly, the same principle applies: it should be less than or equal to max.

Hmm, I am not sure. It will work with larger values too

9il:
see also https://github.com/libmir/mir-random/pull/18, it is required for ushort and ubyte types

WebDrake:
> Hmm, I am not sure. It will work with larger values too

It appears to be applied as a compile-time constraint in the C++11 standard library version.  Try instantiating e.g.:

```C++
std::mersenne_twister_engine<
        UIntType, w, 624, 397, 31,
        0x9908b0df, 11, 0xffffffff, 7,
        0x9d2c5680, 15,
        0xefc60000, 18, (1uL << 32) - 1> gen;
```

... and then try instead:

```C++
std::mersenne_twister_engine<
        UIntType, w, 624, 397, 31,
        0x9908b0df, 11, 0xffffffff, 7,
        0x9d2c5680, 15,
        0xefc60000, 18, (1uL << 32)> gen;
```

PR #18, opened 2017-01-07
make `index` public, change its type to ptrdiff_t

9il:


source/mir/random/engine/mersenne_twister.d
```diff
@@ -58,11 +58,12 @@ struct MersenneTwisterEngine(UIntType, size_t w, size_t n, size_t m, size_t r,
     /// The default seed value.
     enum UIntType defaultSeed = 5489;
 
-    private UIntType _z;
     /++
     Current reversed payload index with initial value equals to `n-1`
     +/
-    private UIntType index = void;
+    ptrdiff_t index = void;
```
WebDrake: Why `ptrdiff_t` and why does it need to be public?

source/mir/random/engine/mersenne_twister.d
```diff
@@ -105,7 +106,7 @@ struct MersenneTwisterEngine(UIntType, size_t w, size_t n, size_t m, size_t r,
         // them separately in sequence, the variables
         // are kept 'hot' in CPU registers, allowing
         // for significantly faster performance.
-        sizediff_t index = cast(size_t)this.index;
+        sizediff_t index = this.index;
```
WebDrake: Small note: it would be preferable to rename the struct field to avoid variable name shadowing here.

9il: shadowing is what i want here

source/mir/random/engine/mersenne_twister.d
9il: hmm, yep, `size_t` would be better better.

9il: https://github.com/libmir/mir-random/commit/f127723463d2b8ec8096b2e3bb11ddd72d924866

9il: It is public to allow user to track Mt state. User may want to initialise it. Mir Random does not have tempalte range API, but we provide an ability for users to do what they want without copy-pasting code

WebDrake: > It is public to allow user to track Mt state. User may want to initialise it.

That seems an odd use-case, to be honest.  As far as initialization is concerned it might be better to provide a means for the user to seed or initialize the generator based on explicit state variables (with clear "use at your own risk" documentation) rather than breaking encapsulation in this way.  (It's not obvious to me that, in your implementation, it could ever be correct to set the index value alone without also ensuring there is a valid `data` array in place.)

As far as _getting_ is concerned, it might be preferable to provide a property method rather than allowing public access to the underlying variable.  Also, note that this implementation's copy of "index" is exactly the opposite of what the index would be in most other implementations of Mersenne Twister, so it might lead to some misleading comparisons if the user is not strongly aware of the implementation details.

source/mir/random/engine/mersenne_twister.d
WebDrake: > shadowing is what i want here

OK, accepted, but can you explain why?  I don't see any benefits to the matching names, and it looks like it could readily be a source of regressions in future if someone editing the code does not take care.

9il: struct methods accept the struct by reference. So I want compiler to do one load and one store of index per opCall. Overriding allows to force to use `this.` to get `index` of struct.

9il:
> That seems an odd use-case, to be honest.

This is what i miss in Phobos. MT is POD struct. Also, Mir libraries aimed to be user-friendly for user plugins. User may want to write initialization function like he want without any API constraints.

source/mir/random/engine/mersenne_twister.d
WebDrake: I'm not sure I follow your argument here.  I'm only concerned about the _naming_ of the local `index` variable being different from the _naming_ of the struct field it takes the value of.

WebDrake:
> This is what i miss in Phobos. MT is POD struct. Also, Mir libraries aimed to be user-friendly for user plugins. User may want to write initialization function like he want without any API constraints.

Ah, I think the context I was missing was that I'd forgotten `data` is already public.  So this change just makes public generator state (in its entirety) the norm.

I accept the use-case, but I wonder whether there are better/safer ways you could achieve it than just making the state public.  Anyway, your library, your use-cases, your call ;-)

PR #24, opened 2017-09-05
add multidimensional distributions

krox:
In my work I often need multidimensional distributions, so I added the two most basic ones:
- uniform distribution of points on a sphere: `x[0]^^2 + .. + x[$-1]^^2 == 1`
- uniform distribution of points on a standard simplex: `x[i] >= 1` and `x[0] + .. + x[$-1] == 1`

The dimension is given as a constructor-argument for `SphereVariable` resp. `SimplexVariable`, and they use an internal buffer so that subsequent calls dont use further allocations.

Thoughts?


jmh530:
You might consider adding an issue for other multivariate distributions that should get added. The most common that I use are Multivariate Normal and Wishart/Inverse Wishart. Multinomial and Dirichlet are also common.

krox:
Okay, I added Dirichlet, that was easy. For Multivariate-Normal one needs to compute the Cholesky decomposition of the covariance matrix. I would assume introducing a dependency on some linear-algebra package is a no-go so I will probably write one myself; should be straight-forward.

But I am not familiar with generating Wishart and Multinomial variables (apart from naive brute-force of course which is inefficient). So I probably leave that for someone else to do.

jmh530:
@krox Cholesky decomposition seems like something perfect to add to `lubeck` (and you really can't do multivariate normal without it). 

There is a lot of statistics that is much easier to express with linear algebra, so I can't imagine this would be the only time this comes up. Personally, I don't see an issue with `mir-random` depending on `mir-glas` (or `lubeck`), but it's not up to me (check with Ilya?).

9il:
@krox thank you for the interesting PR.

Please make a new module, say `ndvariable`. Feel free to propose another module name. Do not forget:
1. to add you as an author
2. to add links to the wiki or other web-pages where user can read more information about the distribution.
3. to add links to most important articles that was used during this work.

API notes:
1. If a function, a method, or a constructor can be a template (even with empty template parameters), then it should be a template. (for better inlining)
2. Please do not store `buf` in the variables. Because:
   - The previous returned `buf` will be invalidated by the new one.
   - It requires memory allocation during variable construction
   - It prevent to use Mir in BetterC programs (D without DRuntime).
 Instead a new API solution should be proposed. 
3. Possible API alternative for `buf`: `void opCall(G, R)(G gen, R result)` where result is something array-like (please do not use isRandomAccessRange and friends, just an unittest with array, should be fine)
4. Please don't allocate Gamma variable array in the constructor. As you can see, GammaVarible construction does not contain any computations. So, it is better to store `alpha` range and construct GammaVariable in the loop on the fly. Please make `Dirichlet` distribution generic for alpha: `struct DirichletVariable(T, AlphaParams = T[])`

Cholesky notes:
You can the low-level D binding for LAPACK, BLAS, and CBLAS:
https://github.com/libmir/mir-lapack/tree/master/source

jmh530:
@krox Refer to dpotrf_ in 
https://github.com/libmir/lapack/blob/master/source/lapack/lapack.d
for cholesky.

9il:
About `buf`s:

Lets make them 1d-ndslices.
```d
    void opCall(G, SliceKind kind, Iterator)(ref G gen, Slice!(kind, [1], Iterator) result)
        if (isSaturatedRandomEngine!G)
    {
        assert(result.length);
        ...
    }
```

Please use `mir.ndslice.sorting` instead of `std.algorithm`

9il:
Please use KBN summation from http://docs.algorithm.dlang.io/latest/mir_math_sum

9il:
> So I probably leave that for someone else to do.

Its fine. Three distributions are already good start for the new module

krox:
I adressed your concers:
- separate module that does not rely on Druntime
- result is supplied by the user either as `T[]` or `Slice!(...)`
- type of parameter-array is templated (without stringent template-constraints)

other notes:
- meaning of `dimension` of sphere/simplex is now mathematically correct. I.e. a circle is one-dimensional even though a point is described by two numbers. Might be confusing to a non-mathematician but the unittest/example of `SphereVariable` should make it clear I think.
- I added Cholesky myself. It's only ~20 lines of code, I think that does not warrant the inclusion of a full linear algebra library. Might change of course if we need more advanced stuff in the future.
- I don't know how to compile the documentation locally (a simple `cd doc; make` doesn't do the trick), so someone should check that it builds okay.

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+        assert(result.length == dim + 1);
+        opCall(gen, result.sliced);
+    }
+
+    ///
+    void opCall(G, SliceKind kind, Iterator)(ref G gen, Slice!(kind, [1], Iterator) result)
+        if (isSaturatedRandomEngine!G)
+    {
+        assert(result.length == dim + 1);
+        for(size_t i = 0; i <= dim; ++i)
```
9il: Please use the following loop style
```d
auto summator = Summator!(T, Summation.kbn)(0);
for(auto r = result; r.front; r.popFront)
{
    auto a = r.front = norm(gen);
    summator += a * a;
}
result[] /= summator.sum.sqrt;
```

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+    {
+        assert(result.length == dim + 1);
+        opCall(gen, result.sliced);
+    }
+
+    ///
+    void opCall(G, SliceKind kind, Iterator)(ref G gen, Slice!(kind, [1], Iterator) result)
+        if (isSaturatedRandomEngine!G)
+    {
+        assert(result.length == dim + 1);
```
9il: please remove assert

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+    +/
+    this(size_t dim)
+    {
+        this.dim = dim;
+    }
+
+    ///
+    void opCall(G)(ref G gen, T[] result)
+    {
+        assert(result.length == dim + 1);
```
9il: please remove assert

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+$(Uniform distribution on a sphere).
+Returns: `X ~ 1` with `X[0]^^2 + .. + X[$-1]^^2 = 1`
++/
+struct SphereVariable(T)
+    if (isFloatingPoint!T)
+{
+    ///
+    enum isRandomVariable = true;
+
+    private size_t dim;
```
9il: please remove the variable `dim`

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+    private NormalVariable!T norm;
+
+    /++
+    Params:
+        dim = dimension of the sphere
+    +/
+    this(size_t dim)
```
9il: Please remove this constructor, variable `private size_t dim;`.

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+Returns: `X ~ 1` with `X[i] >= 0` and `X[0] + .. + X[$-1] = 1`
++/
+struct SimplexVariable(T)
+    if (isFloatingPoint!T)
+{
+    ///
+    enum isRandomVariable = true;
+
+    import mir.ndslice.sorting;
+    private size_t dim;
```
9il: please remove the the variable `dim`

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+
+    /++
+    Params:
+        dim = dimension of the simplex
+    +/
+    this(size_t dim)
```
9il: please remove the constructor

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+    {
+        this.dim = dim;
+    }
+
+    ///
+    void opCall(G)(ref G gen, T[] result)
+    {
+        assert(result.length == dim + 1);
```
9il: just `assert(result.length);`

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+        opCall(gen, result.sliced);
+    }
+
+    ///
+    void opCall(G, SliceKind kind, Iterator)(ref G gen, Slice!(kind, [1], Iterator) result)
+        if (isSaturatedRandomEngine!G)
+    {
+        assert(result.length == dim + 1);
```
9il: ditto

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+
+        for(size_t i = 0; i < dim; ++i)
```
9il: Indexing may be expensive operation for some generic slices.
```d
for(auto r = result[0 .. $ - 1]; r.front; r.popFront)
{
    r.front = gen.rand!T.fabs;
}
...```

File: source/mir/random/ndvariable.d [hunk 11]
```diff
@@ -0,0 +1,293 @@
...
+/++
+$(Multivariate normal distribution).
++/
+struct MultivariateNormalVariable(T, MuParams = T[], SigmaParams = ContiguousMatrix!T)
+    if(isFloatingPoint!T)
+{
+    ///
+    enum isRandomVariable = true;
+
+    private size_t dim;
```
9il: ditto, we do not need this variable.

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+        assert(result.length == dim + 1);
+        opCall(gen, result.sliced);
+    }
+
+    ///
+    void opCall(G, SliceKind kind, Iterator)(ref G gen, Slice!(kind, [1], Iterator) result)
+        if (isSaturatedRandomEngine!G)
+    {
+        assert(result.length == dim + 1);
+        for(size_t i = 0; i < result.length; ++i)
```
9il: ditto, please update loop with range like API and ther fly KBN summation

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+{
+    alias dotm = reduce!"a - b * c"; // note the `-`
+    assert(m.length!0 == m.length!1);
+
+    /* this is a straight-forward implementation of the Cholesky-Crout algorithm
+    from https://en.wikipedia.org/wiki/Cholesky_decomposition#Computation */
+    for(size_t i = 0; i < m.length!0; ++i)
+    {
+        for(size_t j = 0; j < i; ++j)
+            m[i,j] = dotm(m[i,j], m[i,0..j], m[j,0..j]) / m[j,j];
```
9il: please use this style `m[i,0..j]` -> `m[i, 0 .. j]`

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+ +/
+private bool cholesky(SliceKind kind, Iterator)(Slice!(kind, [2], Iterator) m)
+    if(isFloatingPoint!(DeepElementType!(typeof(m))))
+{
+    alias dotm = reduce!"a - b * c"; // note the `-`
+    assert(m.length!0 == m.length!1);
+
+    /* this is a straight-forward implementation of the Cholesky-Crout algorithm
+    from https://en.wikipedia.org/wiki/Cholesky_decomposition#Computation */
+    for(size_t i = 0; i < m.length!0; ++i)
```
9il: something like that, to catche indexing operations:
```d
foreach(size_t; 0 .. m.length!0)
{
    auto mi = m[i];
    foreach(size_t j; 0 .. i)
    {
        auto mj = m[j]l
        mi[j] = (mi[j] - dot(mi[0 .. j], mj[0 .. j])) / mj[j];
    }
}
```
Also please add a private common `dot` functions and use it instead of all `reduce` in this module

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+
+    private size_t dim;
+    private AlphaParams alpha;
+
+    /++
+    Params:
+        alpha = (array of) concentration parameters
+    Constraints: `alpha[i] > 0`
+    +/
+    this(AlphaParams alpha)
```
9il: BTW, please make all constructors generic too: `this()(AlphaParams alpha) `

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+
+    ///
+    void opCall(G)(ref G gen, T[] result)
+    {
+        assert(result.length == dim);
+        opCall(gen, result.sliced);
+    }
+
+    ///
+    void opCall(G, SliceKind kind, Iterator)(ref G gen, Slice!(kind, [1], Iterator) result)
```
9il: ditto about loop, and private common `dot` instead of reduce

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+    memory allocation. Furthermore it is assumed to be a symmetric matrix, but
+    only the lower/left half is actually accessed.
+
+    Params:
+        mu = mean vector
+        sigma = covariance matrix
+
+    Constraints: sigma has to be positive-definite
+    +/
+    this(MuParams mu, SigmaParams sigma)
```
9il: `this(MuParams mu, SigmaParams sigma, bool computeCholesky = true)`

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,293 @@
...
+            return false;
+        m[i,i] = sqrt(m[i,i]);
+    }
+    return true;
+}
+
+/++
+$(Multivariate normal distribution).
++/
+struct MultivariateNormalVariable(T, MuParams = T[], SigmaParams = ContiguousMatrix!T)
```
9il: `MuParams = const(T)[]`

9il:
thanks for the update

source/mir/random/ndvariable.d
jmh530: I like this suggestion as sometimes one passes the Cholesky. It might make sense to re-name `private SigmaParams sigma;` to `private SigmaParams cholSigma;`

I would also recommend adding the following constructor
`this(SigmaParams sigma, bool computeCholesky = true)`
that assumes that mu is a zero vector.

krox:
I did:
- made imports local in apropriate scopes. Except for `mir.ndslice` which has to be global (beacuse it appears in the the declaration of various types, which cannot see their own bodies)
- removed the `dim` fields, even though I kinda liked them. Now some types (Dirichlet/MultiNormal) have a fixed dimension and some (Sphere/Simplex) do not, which seems inconsistent to me. So you are right of course that they are redunant so I'm fine with removing them to slightly remove overhead.
- more general Cholesky-constructor. But I inverted the flag, i.e. `bool cholAlreadyDone = false` which feels more natural to me (and consistent with a similar parameter in `PiecewiseConstantVariable`

other things I don't really agree with:
- Empty template `()`: This can avoid emission of any code in case some function is not used at all. But here, everything is inside templated struct already, so thats superfluous.
- sophisticated summation methods: Nice algorithms but they cost serious performance (doubles or triples number of instructions and effectively disables vectorization I think. Though I might be wrong). So I don't think they should be used by default (afterall, we don't use them inside linear-algebra computations even though they could in principle increase accuracy).

Still, thats just my two cents, so feel free to do any further changes you like. :)

krox:
whoops, I did not mean to close it^^

source/mir/random/ndvariable.d
```diff
@@ -0,0 +1,289 @@
...
+        if(!chol && !cholesky(sigma))
+            assert(false, "covariance matrix not positive definite");
+
+        this.mu = mu;
+        this.sigma = sigma;
+        this.norm = NormalVariable!T(0, 1);
+    }
+
+    /++ ditto +/
+    this(SigmaParams sigma, bool chol = false)
```
jmh530: You might add a unittest for the alternate constructor.

9il:
> Empty template ():
> This can avoid emission of any code in case some function is not used at all.

Exactly, this is what we need. Everything that may not be used and can be generic should be generic.

> But here, everything is inside templated struct already, so thats superfluous.

No, you may instant template (just to check the type) but you do not need to instant its methods and functions. Compilation takes time.

9il:
> sophisticated summation methods: Nice algorithms but they cost serious performance (doubles or triples number of instructions and effectively disables vectorization I think. Though I might be wrong). So I don't think they should be used by default (afterall, we don't use them inside linear-algebra computations even though they could in principle increase accuracy).

No, the cost almost nothing comparing with random number generation. Please use proposed loops pattern with summation on the fly.

9il:
And please update loops with more dlang/ndslice style as was proposed.

9il:
Will fix it myself

PR #26, opened 2017-10-09
Use arc4random_buf for unpredictableSeed when known to be good

n8sh:
Also use it for genRandomNonBlocking. On Mac OS X /dev/random does not
block so also use arc4random_buf for genRandomBlocking.


9il:
LGTM

PR #28, opened 2017-10-10
Fix isSaturatedRandomEngine when opCall is a function template

n8sh:
std.traits.ReturnType!T doesn't work if T.opCall is a function template.
This meant that isSaturatedRandomEngine and a few other things would
not compile when given the mir.random.engine.pcg generators as template
arguments.


9il:
Thank you!

PR #29, opened 2017-10-10
Allow seeding MersenneTwisterEngine with arrays/ranges

n8sh:
Behavior is like the reference implementation's init_by_array,
not like Phobos std.random.MersenneTwisterEngine's seed(Range).

### DIFF
--- a/source/mir/random/engine/mersenne_twister.d
+++ b/source/mir/random/engine/mersenne_twister.d
@@ -12,18 +12,24 @@ import mir.ndslice.slice : Slice, SliceKind, Contiguous;
 import std.range.primitives : isInputRange, ElementType;
 
 /*
-For betterC define our own private pureMalloc, etc.
-to avoid depending on druntime core.memory.
-*/
+ Can't use pureMalloc from core.memory because no druntime in betterC.
+ Can't do the following because it doesn't build with LDC:
+
 extern (C) private @nogc nothrow pure
 {
     //We can consider this malloc to effectively be pure
     //because it is private and we immediately abort the
     //program if it fails.
-    pragma(mangle, "malloc") void* pureMalloc(size_t) @trusted;
-    pragma(mangle, "realloc") void* pureRealloc(void* ptr, size_t size); @system
-    pragma(mangle, "free") void* pureFree(void* ptr) @system;
+    pragma(mangle, "malloc") void* mtPureMalloc(size_t) @trusted;
+    pragma(mangle, "realloc") void* mtPureRealloc(void* ptr, size_t size); @system
+    pragma(mangle, "free") void* mtPureFree(void* ptr) @system;
 }
+
+ So we have to use impure malloc/free. We have to do this with both
+ DMD and LDC, because otherwise if a user called certain methods from
+ `pure` code his own program would fail to build with LDC.
+*/
+private import core.stdc.stdlib : free, malloc, realloc;
 
 /++
 The $(LUCKY Mersenne Twister) generator.
@@ -215,7 +221,7 @@ struct MersenneTwisterEngine(UIntType, size_t w, size_t n, size_t m, size_t r,
             immutable length = range.length;
             if (length > size_t.max / UIntType.sizeof)
                 assert(0, "MersenneTwister: input range too large to be converted to array!");
-            void* raw_buffer = pureMalloc(length * UIntType.sizeof);
+            void* raw_buffer = (() @trusted => malloc(length * UIntType.sizeof))();
             if (raw_buffer is null)
                 assert(0, "MersenneTwister: malloc failed!");
             UIntType[] buffer = (() @trusted => (cast(UIntType*) raw_buffer)[0 .. length])();
@@ -226,7 +232,7 @@ struct MersenneTwisterEngine(UIntType, size_t w, size_t n, size_t m, size_t r,
             else
             {
                 enum bool mustManuallyFreeBuffer = false;
-                scope(exit) (() @trusted => pureFree(raw_buffer))();
+                scope(exit) (() @trusted => free(raw_buffer))();
             }
             foreach (ref e; buffer)
             {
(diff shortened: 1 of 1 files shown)


9il:
@n8sh Could you please rework it to
three generic methods:

The first one accepts `scope Slice!(Contiguous, [1], const(UInt)*)`;

the second one accepts an `scope const (UInt)[]`, applies `.sliced`, and passes it to the first one;

the third accepts an input range (no need to be forward), fills a local buffer from it and passes this buffer to the second one method.

n8sh:
@9il I'm willing to make the change but there may be something I am misunderstanding.

> The first one accepts scope Slice!(Contiguous, [1], const(UInt)*);
> the second one accepts an scope const (UInt)[], applies .sliced, and passes it to the first one;

The code I have written that uses `scope Slice!(Contiguous, [1], const(UIntType)*)` looks like this:
```
        size_t i = n - 2;
        size_t j = 0;
        immutable key_length = slice.length;
        const(UIntType)[] field = slice.field; // <-- ???
        foreach_reverse(_; 0 .. (n > key_length ? n : key_length))
        {
            data[i] = (data[i] ^ ((data[i+1] ^ (data[i+1] >> (w - 2))) * f2))
                + cast(UIntType)(field[j] + j);
            static if (max != UIntType.max)
                data[i] &= max;
            if (--i > n)
            {
                data[$ - 1] = data[0];
                i = n - 2;
            }
            if (++j >= key_length)
                j = 0;
        }
```
Is it poor style to use `slice.field` to get `const(UIntType)[]` here? It seems odd to wrap `const(UIntType)[]` in a Slice struct then immediately unwrap it.

9il:
Just uses slice instead, it should just work. `slice.length` is zero cost.
```d
        import mir.utility: min;
        size_t i = n - 2;
        size_t j = 0;
        foreach_reverse(_; 0 .. min(slice.length, n))
        {
            data[i] = (data[i] ^ ((data[i+1] ^ (data[i+1] >> (w - 2))) * f2))
                + cast(UIntType)(slice[j] + j);
            static if (max != UIntType.max)
                data[i] &= max;
            if (--i > n)
            {
                data[$ - 1] = data[0];
                i = n - 2;
            }
            if (++j >= slice.length)
                j = 0;
        }
```

9il:
BTW, could you please slightly optimise the loop?
```d
        import mir.utility: min;
        size_t j = 0;
        immutable len = min(slice.length, n) - 1;
        foreach_reverse(i; 0 .. len) // do not handle tail
        {
            data[i] = (data[i] ^ ((data[i+1] ^ (data[i+1] >> (w - 2))) * f2))
                + cast(UIntType)(slice[j] + (len - 1 - i));
            static if (max != UIntType.max)
                data[i] &= max;
        }
       // handle tail conditions here
```

source/mir/random/engine/mersenne_twister.d
```diff
@@ -90,6 +106,198 @@ struct MersenneTwisterEngine(UIntType, size_t w, size_t n, size_t m, size_t r,
...
+                j = 0;
+        }
+
+        foreach_reverse(_; 0 .. n-1)
+        {
+            data[i] = (data[i] ^ ((data[i+1] ^ (data[i+1] >> (w - 2))) * f3))
+                - cast(UIntType)(n - (i + 1)); /* non linear */
+            static if (max != UIntType.max)
+                data[i] &= max;
+            if (--i > n)
```
9il: the same here

source/mir/random/engine/mersenne_twister.d
```diff
@@ -90,6 +106,198 @@ struct MersenneTwisterEngine(UIntType, size_t w, size_t n, size_t m, size_t r,
...
+    Constructs a MersenneTwisterEngine object.
+
+    This method copies the values from the range into a temporary buffer.
+    If you can pass an array or slice directly that is more efficient.
+
+    A range that is detected as infinite by `std.range.primitives.isInfinite`
+    will be treated as though its length were `n`.
+    +/
+    this(Range)(scope Range range)
+    if (isInputRange!Range && is(Unqual!(ElementType!Range) == UIntType))
```
9il: this should be three line of code function:
```
import std.algorithm.mutation: copy;
import std.range: take;
this(buffer[0 .. (range.take(n).copy(buffer[]).length].sliced);
```

n8sh:
Including private pure versions pureMalloc and pureFree builds with DMD but not LDC due to this issue:

>Error: Function type does not match previously declared function with the same mangled name: free

So on LDC it looks like the only option is to use impure `free`. This might cause slight annoyance for users because if they call the InputRange constructor from `pure` code then _their_ program will compile with DMD but not on LDC. So if we're not using the GC and not using druntime, then I think the code will have to use impure `malloc`/`free` with all compilers.

source/mir/random/engine/mersenne_twister.d
n8sh: Wouldn't that introduce a dependency on druntime?

9il: it would not

9il:
> Including private pureMalloc and pureFree builds with DMD but not LDC due to this issue:
> 
> Error: Function type does not match previously declared function with the same mangled name: free
> So on LDC it looks like the only option is to use impure free. This might cause slight annoyance for users because if they call the InputRange constructor from pure code then their program will compile with DMD but not on LDC. So if we're not using the GC and not using druntime, then I think the code will have to use impure malloc/free with all compilers.

1. You do not need to allocate memory at all because the max used length us know at compile time.
2. Just use this three LOC

9il:
If so, lets just remove the method for ranges and use existing for array and slice only

n8sh:
>Do we really need to use entire array if it length is larger then N?

If we want it to match the reference implementation.

>If so, lets just remove the method for ranges and use existing for array and slice only

Alright.

PR #31, opened 2017-10-22
Fix increment of PCG unique_stream

n8sh:
The increment is supposed to be based on the address of the object
but that is not what the code did.


9il:
Thank you both!

PR #32, opened 2017-10-22
Make bump() static for PCG oneseq_stream and no_stream

n8sh:


thewilsonator:
I'm not sure that the inconsistency in the interface is worth it, considering the benefit of adding static is very marginal and should be dealt with the same by the optimiser. What does the C++ implementation do?

@9il thoughts?

n8sh:
C++ implementation doesn't make it static.

Since the function is private it probably doesn't matter whether the interface is consistent, but I agree it would be fine to leave this to the compiler.

The main reason I made the change is that
```
    static assert (pcg32_fast.sizeof == ulong.sizeof);
```
was failing, and making sure that the `bump()` function was static addressed the reason for this.

9il:
This is interface question. Does PCG specify anything about non default increments or only default increment is considered?

thewilsonator:
@n8sh are you sure it wasn't `enum Uint increment = 0;` that changed it?

n8sh:
(Although the reason the C++ version doesn't make it static might also be because it's harder to write checks like that in C++.)

@9il 
> Does PCG specify anything about non default increments or only default increment is considered?

It addresses the question. Whether the increment is changeable depends on the stream type.

specific_stream: can be changed
oneseq_stream: cannot be changed (constant)
no_stream: cannot be changed (constant, zero)
unique_seq: cannot be changed (based on address)

n8sh:
@thewilsonator 
>@n8sh are you sure it wasn't enum Uint increment = 0; that changed it?

That is what what changed it. What I meant was the way I found it was by looking for reasons why `bump` couldn't be static and changing things until it could.

n8sh:
`pcg32_once_insecure` should have non-static bump and `pcg32_oneseq_once_insecure` should have static bump. I'll add them to the unit test and re-push it.

PR #39, opened 2017-10-25
Use Daniel Lemire’s fast alternative to modulo reduction

n8sh:
[Full explanation on Daniel Lemire's blog.](https://lemire.me/blog/2016/06/27/a-fast-alternative-to-the-modulo-reduction/) [Another entry covers bias correction when using this method.](https://lemire.me/blog/2016/06/30/fast-random-shuffling/)

If our `gen()` produces `ulong` and we want to convert this to a `uint` in range `0 .. m`, instead of
```
(cast(uint) gen()) % m
```
we can do:
```
cast(uint) (((gen() >>> 32) * m) >>> 32)
```
which replaces a modulo operation with a multiplication and some shifts.

I put a benchmark for this at [bench/randindex_bench.d](https://github.com/n8sh/mir-random/blob/9fe19f33766cdce7baa00b3b024d5c0d2e5bd29b/bench/randindex_bench.d). The performance gain was minor, but the result persisted when running the benchmark repeatedly.

----
**EDIT:**
As @9il points out below, the original benchmark was flawed due to the compiler being too clever and optimizing for a constant modulus. After working to prevent this, the speed advantage of the new version [increased](https://github.com/n8sh/mir-random/blob/2e90600bd905d71653a0735058b296fbe1061222/bench/randindex_bench.d).

### DIFF
--- a/bench/randindex_bench.d
+++ b/bench/randindex_bench.d
@@ -1,0 +1,151 @@
+#!/usr/bin/env dub --build=release-nobounds --compiler=ldmd2 -v --single
+/+ dub.json: {
+    "name":"randindex_bench",
+    "dependencies": {
+        "mir-random":{"path": "../"}
+    }
+} +/
+import mir.random.engine : isSaturatedRandomEngine, EngineReturnType;
+import mir.random.engine.mersenne_twister : Mt19937, Mt19937_64;
+import mir.random.engine.pcg : pcg32_oneseq, pcg64_oneseq_once_insecure;
+import mir.random.engine.xorshift : Xoroshiro128Plus, Xorshift1024StarPhi;
+import mir.utility : min, max;
+
+import std.traits: isUnsigned, Unqual;
+
+import std.stdio, std.datetime, std.conv;
+
+/*
+Sample results on Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
+
+Benchmarks for MersenneTwisterEngine!(ulong, 64LU, 312LU, 156LU, 31LU, 13043109905998158313LU, 29LU, 6148914691236517205LU, 17LU, 8202884508482404352LU, 37LU, 18444473444759240704LU, 43LU, 6364136223846793005LU):
+randIndexV1: 2.32964 * 10 ^^ 8 calls/s
+randIndexV2: 2.48911 * 10 ^^ 8 calls/s
+
+Benchmarks for XorshiftStarEngine!(ulong, 1024u, 31u, 11u, 30u, 11400714819323198483LU, ulong):
+randIndexV1: 3.73308 * 10 ^^ 8 calls/s
+randIndexV2: 3.78072 * 10 ^^ 8 calls/s
+
+Benchmarks for Xoroshiro128Plus:
+randIndexV1: 6.09292 * 10 ^^ 8 calls/s
+randIndexV2: 7.22022 * 10 ^^ 8 calls/s
+
+Benchmarks for PermutedCongruentialEngine!(rxs_m_xs_forward, cast(stream_t)2, true):
+randIndexV1: 4.40771 * 10 ^^ 8 calls/s
+randIndexV2: 3.87409 * 10 ^^ 8 calls/s <-- check this out
+
+--------
+The performance appeared to be worse for the PCG for some reason.
+I tried special-casing the algorithm for PCGs but that didn't help,
+leading me to suspect that something else was going on.
+
+To investigate I ran another test with randIndexV2 being run again
+with the label "randIndexV3".
+--------
+
+Benchmarks for MersenneTwisterEngine!(ulong, 64LU, 312LU, 156LU, 31LU, 13043109905998158313LU, 29LU, 6148914691236517205LU, 17LU, 8202884508482404352LU, 37LU, 18444473444759240704LU, 43LU, 6364136223846793005LU):
+randIndexV1: 2.28115 * 10 ^^ 8 calls/s
+randIndexV2: 2.41036 * 10 ^^ 8 calls/s
+randIndexV3: 2.41619 * 10 ^^ 8 calls/s
+
+Benchmarks for XorshiftStarEngine!(ulong, 1024u, 31u, 11u, 30u, 11400714819323198483LU, ulong):
+randIndexV1: 3.64133 * 10 ^^ 8 calls/s
+randIndexV2: 3.76471 * 10 ^^ 8 calls/s
+randIndexV3: 3.75059 * 10 ^^ 8 calls/s
+
+Benchmarks for Xoroshiro128Plus:
+randIndexV1: 6.03774 * 10 ^^ 8 calls/s
+randIndexV2: 7.08592 * 10 ^^ 8 calls/s
+randIndexV3: 7.02371 * 10 ^^ 8 calls/s
+
+Benchmarks for PermutedCongruentialEngine!(rxs_m_xs_forward, cast(stream_t)2, true):
+randIndexV1: 4.28266 * 10 ^^ 8 calls/s
+randIndexV2: 3.75587 * 10 ^^ 8 calls/s
+randIndexV3: 4.25532 * 10 ^^ 8 calls/s <--- check this out
+
+
+We can see that for PCG, it isn't that one method is faster than the
+other, but something else is going on.
+*/
+
+T randIndexV1(T, G)(ref G gen, T m)
+    if(isSaturatedRandomEngine!G && isUnsigned!T)
+{
+    assert(m, "m must be positive");
+    T ret = void;
+    T val = void;
+    do
+    {
+        //val = gen.rand!T;
+        val = cast(T) gen();
+        ret = val % m;
+    }
+    while (val - ret > -m);
+    return ret;
+}
+
+T randIndexV2(T, G)(ref G gen, T m)
+    if (isSaturatedRandomEngine!G && isUnsigned!T
+        && (T.sizeof * 2) <= (EngineReturnType!G).sizeof)
+{
+    //https://lemire.me/blog/2016/06/30/fast-random-shuffling/
+    alias R = Unqual!(EngineReturnType!G);
+    enum rshift = (R.sizeof - T.sizeof) * 8;
+    auto randombits = gen() >>> rshift;
+    auto multiresult = randombits * m;
+    auto leftover = cast(T) multiresult;
+    if (leftover < m)
+    {
+        immutable threshold = -m % m ;
+        while (leftover < threshold)
+        {
+            randombits =  gen() >>> rshift;
+            multiresult = randombits * m;
+            leftover = cast(T) multiresult;
+        }
+    }
+    return cast(T) (multiresult >>> rshift);
+}
+
+void main(string[] args)
+{
+    import std.meta : AliasSeq;
+    uint s = 0;
+    foreach (PrngType; AliasSeq!(Mt19937_64, Xorshift1024StarPhi, Xoroshiro128Plus, pcg64_oneseq_once_insecure))
+    {
+        writeln("\nBenchmarks for ", PrngType.stringof, ":");
+        enum seed = PrngType.max / 2;
+        auto gen = PrngType(seed);
+        enum ulong count = 800_000_000;
+        StopWatch sw;
+        sw.start;
+        foreach(_; 0 .. min(count / 2, 200_000_000u)) //boost CPU
+        {
+            s += gen.randIndexV1(cast(uint)100);
+            s += gen.randIndexV2(cast(uint)100);
+        }
+        sw.stop;
+        sw.reset;
+        gen.__ctor(seed);
+        sw.start;
+        foreach(_; 0..count)
+            s += gen.randIndexV1(cast(uint)100);
+        sw.stop;
+        writefln("randIndexV1: %s * 10 ^^ 8 calls/s", double(count) / sw.peek.msecs / 100_000);
+        sw.reset;
+        gen.__ctor(seed);
+        sw.start;
+        foreach(_; 0..count)
+            s += gen.randIndexV2(cast(uint)100);
+        sw.stop;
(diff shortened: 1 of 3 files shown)


bench/randindex_bench.d
```diff
@@ -0,0 +1,152 @@
...
+        ret = val % m;
+    }
+    while (val - ret > -m);
+    return ret;
+}
+
+T randIndexV2(T, G)(ref G gen, T m)
+    if (isSaturatedRandomEngine!G && isUnsigned!T
+        && (T.sizeof * 2) <= (EngineReturnType!G).sizeof)
+{
```
9il: please mark both with `pragma(inline, false)`. Current benchmark is not fair because LDC is clever enough to optimise algorithm for `100`

bench/randindex_bench.d
```diff
@@ -0,0 +1,152 @@
...
+    //https://lemire.me/blog/2016/06/30/fast-random-shuffling/
+    alias R = Unqual!(EngineReturnType!G);
+    enum rshift = (R.sizeof - T.sizeof) * 8;
+    auto randombits = gen() >>> rshift;
+    auto multiresult = randombits * m;
+    auto leftover = cast(T) multiresult;
+    if (leftover < m)
```
9il: after `pragma(inline, false)`, try to replace it with `_expect(leftover < m, false)`, `_expect` is from  `mir.ndslice.internal`

bench/randindex_bench.d
```diff
@@ -0,0 +1,152 @@
...
+        writefln("randIndexV2: %s * 10 ^^ 8 calls/s", double(count) / sw.peek.msecs / 100_000);
+        //sw.reset;
+        //gen.__ctor(seed);
+        //sw.start;
+        //foreach(_; 0..count)
+        //    s += gen.randIndexV2(cast(uint)100);
+        //sw.stop;
+        //writefln("randIndexV3: %s * 10 ^^ 8 calls/s", double(count) / sw.peek.msecs / 100_000);
+    }
+    writeln("Meaningless sum: ", s);
```
9il: Please add a test that shows that generated numbers are uniformly distributed.

source/mir/random/package.d
```diff
@@ -317,16 +317,46 @@ Returns:
...
-    assert(m, "m must be positive");
-    T ret = void;
-    T val = void;
-    do
+    alias R = EngineReturnType!G;
+    version (D_X32)
+        enum bool versionD_X32 = true;//On 64-bit machine but using 32-bit pointers.
+    else
+        enum bool versionD_X32 = false;
+    static if (R.sizeof >= T.sizeof * 2 && (size_t.sizeof >= R.sizeof || versionD_X32))
```
9il: - The original algorithm generates the same amount of bits as the result has. It do not need to generate more. - So, it should be just `gen.rand!T`  instead of `gen() >>> rshift` : `rand` will chose lowest or highest bits depends on the engine.
- We only need to check that `multiresult` has at least twice more bits than R.

The problem for this optimisation is in practice `T` is `size_t`. In other hand `x86` and `x86_64` generates extended multiplication results stored in two registers: 32x32->64 and 64x64->128 (x86_64).

- ubyte and ushort T: `randombits * m;` (the result will be 32 bit because of language semantics)
- uint T needs `randombits * cast(ulong) m;` and compiler will do its best job for all targets.
- ulong T needs to use LLVM IR if no `ucent` or `randombits * cast(ucent) m;` if ucent
- ucent T need to use LLVM IR

LLVM IR
 - https://wiki.dlang.org/LDC_inline_IR
 - https://llvm.org/docs/LangRef.html

9il: - Please add `@safe` attribute for unittests.
- Please addd CTFE unittests. LLVM IR can no be executed compile time, so the old code should be used in CTFE (always for simplicity).

9il: `= void` probably unsafe, but I am not sure

n8sh:
>please mark both with pragma(inline, false). Current benchmark is not fair because LDC is clever enough to optimise algorithm for 100

Ah! That may explain some of the surprises.

In the next day or two I will revise the test and try again.

9il:
Some CPU cache related nitpick for the benchmark: The following code should be before both tests. Also please print results after all tests.
```d
+        foreach(_; 0 .. min(count / 2, 200_000_000u)) //boost CPU
 +        {
 +            s += gen.randIndexV1(cast(uint)100);
 +            s += gen.randIndexV2(cast(uint)100);
 +        }
```

9il:
I just realise that we can force saturated engine to be safe. So please do not care about it in this PR.

9il:
https://github.com/libmir/mir-random/issues/42

n8sh:
>The problem for this optimisation is in practice T is size_t.

I've now added LDC-only code that uses LLVM IR to generate 64-bit size_t. Also, now the method used to reduce the value has nothing to do with `EngineReturnType!G` but only `T`.

n8sh:
>Please add a test that shows that generated numbers are uniformly distributed.

Added.

>Also please print results after all tests.

Added.

@9il I think this is everything.

n8sh:
@9il 
>`= void` probably unsafe, but I am not sure

Because it's not a pointer it is `@safe` for it to have a void initializer according to D.

9il:
needs rebase

source/mir/random/package.d
```diff
@@ -307,6 +321,46 @@ version(mir_random_test) unittest
...
+
+            pragma(inline, true)
+            core.simd.ulong2 mul_128(ulong a, ulong b)
+            {
+                return inlineIR!(`
+                    %a = zext i64 %0 to i128
+                    %b = zext i64 %1 to i128
+                    %ra = mul i128 %a, %b
+                    %rb = bitcast i128 %ra to <2 x i64>
+                    ret <2 x i64> %rb`, core.simd.ulong2)(a, b);
```
9il: Please do not use SIMD here, they makes code slightly more complicated.
`ulong[2]` can be returned. `(a, b, &c, &d)` where ulong c, d; and return `[a, b]`. It is OK to use 64 bit shift, LLVM is clever enough.

9il: In addition you would not need to care Endians

bench/randindex_bench.d
```diff
@@ -0,0 +1,392 @@
...
+    return ret;
+}
+
+version (LDC)
+{
+    //TODO: figure out specific feature flag or CPU versions where 128 bit multiplication works!
+    version (X86_64)
+        private enum bool probablyCanMultiply128 = true;
+    else
+        private enum bool probablyCanMultiply128 = size_t.sizeof >= ulong.sizeof;
```
9il: If I am not wrong, LLVM can always handle 128 / 256 bit multiplication, for targets where it is not supported it will be simulated.

bench/randindex_bench.d
```diff
@@ -0,0 +1,392 @@
...
+                s1 += gen.randIndex!uint(m);
+            s += s1;
+        }
+        sw.stop;
+        writefln("new mir.random.randIndex (potential inlining shenanigans): %s * 10 ^^ 8 calls/s; sum = %d", double(count) / sw.peek.msecs / 100_000, s);
+    }
+
+    {
+        //Uniform distribution check.
+        static struct Counter
```
9il: Good idea!

bench/randindex_bench.d
```diff
@@ -0,0 +1,392 @@
...
+            enum bool isRandomEngine = true;
+            enum uint max = uint.max;
+            uint state;
+            @disable this();
+            @disable this(this);
+            this(uint state) { this.state = state; }
+            uint opCall() { return state++; }
+        }
+        import mir.random.engine;
+        enum uint nbuckets = uint(1u << 16);
```
9il: The problem with this number is that `uint.max  % nbuckets  == 0`. So the while loop is never checked.

bench/randindex_bench.d
```diff
@@ -0,0 +1,392 @@
...
+            foreach (m; modulus_min .. modulus_max)
+                s1 += gen.randIndex!uint(m);
+            s += s1;
+        }
+        sw.stop;
+        writefln("new mir.random.randIndex (potential inlining shenanigans): %s * 10 ^^ 8 calls/s; sum = %d", double(count) / sw.peek.msecs / 100_000, s);
+    }
+
+    {
+        //Uniform distribution check.
```
9il: We should have it in unittests for all unsigned types.

source/mir/random/package.d
```diff
@@ -332,6 +332,44 @@ T rand(T, G)(ref G gen, sizediff_t boundExp = 0)
...
 
+version (LDC)
+{
+    //TODO: figure out specific feature flag or CPU versions where 128 bit multiplication works!
+    version (X86_64)
+        private enum bool probablyCanMultiply128 = true;
+    else
+        private enum bool probablyCanMultiply128 = size_t.sizeof >= ulong.sizeof;
+
+    static if (probablyCanMultiply128 && !is(ucent))
```
9il: Please add 256bit for ucent too. ucent division will be very slow

bench/randindex_bench.d
```diff
@@ -0,0 +1,392 @@
...
+    else static if (ulong.sizeof >= T.sizeof * 2)
+        alias MaybeR = ulong;
+    else static if (is(ucent) && __traits(compiles, {static assert(ucent.sizeof >= T.sizeof * 2);}))
+        mixin ("alias MaybeR = ucent;");
+    else
+        alias MaybeR = void;
+
+    static if (!is(MaybeR == void))
+    {
+        if (!__ctfe)
```
9il: I will do few changes,  please do not force push!

9il:
I will add a extended multiplication to mir-algorithm, this will simplify slightly code here.

9il:
Please use http://docs.algorithm.dlang.io/latest/mir_utility.html#.extMul in `v0.7.0-alpha4`  both for runtime and compile time code. Old algorithm with division can be removed.

bench/randindex_bench.d
n8sh: I thought so too, but travis-ci compilation failed for x86 on Linux.

9il: Yep, mir-algorithm uses common mul algo now. But it should be faster then div anyway

n8sh:
@9il
>// LLVM IR by n8sh

BTW if for some reason you want to put my name on other things you can write "Nathan Sashihara".

9il:
Added you to the xorshift authors. Please add yourself with this PR into the package.

n8sh:
@9il  When compiled by DMD for Windows with --arch=x86_64, `extMul_X86_64` crashes.

```
private ExtMulResult!ulong extMul_X86_64()(ulong a, ulong b)
{
    // pragma(inline, true);
    asm @safe pure nothrow @nogc
    {
        naked;
        mov RAX, RDI;
        mul RSI;
        ret;
    }
}
```

bench/randindex_bench.d
```diff
@@ -152,9 +114,9 @@ T randIndexV2(T, G)(ref G gen, T m)
     else
         alias MaybeR = void;
 
-    static if (!is(MaybeR == void))
+    version (LDC) if (!__ctfe)
```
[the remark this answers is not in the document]
n8sh: The issue is _expect

bench/randindex_bench.d
```diff
@@ -0,0 +1,379 @@
...
+        {
+            alias R = MaybeR;
+            static assert(R.sizeof >= T.sizeof * 2);
+            import mir.ndslice.internal: _expect;
+            //Use Daniel Lemire's fast alternative to modulo reduction:
+            //https://lemire.me/blog/2016/06/30/fast-random-shuffling/
+            R randombits = cast(R) gen.rand!T;
+            R multiresult = randombits * m;
+            T leftover = cast(T) multiresult;
+            if (_expect(leftover < m, false))
```
9il: OK, then `if` should have most probable condition:
```
...
if (leftover >= m)
    return cast(T) (multiresult >>> finalshift);
...
while(...)
```

n8sh:
>Could you please show the output?

Nothing useful:
```
Running .\mir-random-test-library.exe
Program exited with code -1073741819
```

I had to use a debugger to determine where it was crashing.

n8sh:
@9il I've written a tiny program that demonstrates the problem on Windows when compiled with `dmd -m64`:

```
import std.traits;

/++
Return type for $(LREF extMul);
+/
struct ExtMulResult(I)
    if (isIntegral!I)
{
    /// Lower I.sizeof * 8 bits
    I low;
    /// Higher I.sizeof * 8 bits
    I high;

}

version(Windows)
version(D_InlineAsm_X86_64)
private ExtMulResult!ulong extMul_X86_64_Windows()(ulong a, ulong b)
{
    asm @nogc nothrow pure @safe
    {
        naked;
        mov RAX, RCX;
        mul RDX;
        ret;
    }
}

void main(string[] args)
{
    import std.stdio;
    auto x = extMul_X86_64_Windows(123,2);
    //This line "works" although the result is garbage:
    writeln(x);
    //This line crashes:
    writeln("extMul_X86_64_Windows(123,2) = ", extMul_X86_64_Windows(123,2));
}
```

bench/randindex_bench.d
n8sh: Is it really a problem to use _expect in runtime code only? We have a unittest that verifies that CTFE and runtime evaluation produce the same result.

9il:
> mov RAX, RCX;

This is not like in the original code, it was `mov RAX, RSI`

9il:
> EDIT: actually, running this code on a Mac doesn't produce the desired result either, although it doesn't crash. The asm block in extMul_X86_64 seems to be wrong?

This code is wrong. Original has uses RSI register

n8sh:
>This is not like in the original code, it was mov RAX, RSI

>This code is wrong for OSX. Original uses RSI register

My mistake -- I changed it for Windows calling conventions and forgot to change it back before testing on another machine.

9il:
Great addition!

PR #45, opened 2017-10-26
`@safe unittest` for every function in mir.random.algorithm

n8sh:
Includes changes required for `unittest`s to pass.
Requires mir.random.package to have been made `@safe`
Part of ongoing effort https://github.com/libmir/mir-random/issues/40

### DIFF
--- a/source/mir/random/algorithm.d
+++ b/source/mir/random/algorithm.d
@@ -28,7 +28,7 @@ struct RandomField(G, D, T)
     Constructor.
     Stores the pointer to the `gen` engine.
     +/
-    this()(ref G gen, D var) { _gen = &gen; _var = var; }
+    this()(ref G gen, D var) { _gen = (() @trusted => &gen)(); _var = var; }
     ///
     T opIndex()(size_t)
     {
@@ -54,7 +54,7 @@ struct RandomField(G)
     Constructor.
     Stores the pointer to the `gen` engine.
     +/
-    this()(ref G gen) { _gen = &gen; }
+    this()(ref G gen) @trusted { _gen = &gen; }
     ///
     Unqual!(EngineReturnType!G) opIndex()(size_t) { return (*_gen)(); }
 }
@@ -81,7 +81,7 @@ RandomField!G field(G)(ref G gen)
 }
 
 /// Normal distribution
-version(mir_random_test) unittest
+nothrow @safe version(mir_random_test) unittest
 {
     import mir.ndslice: slicedField, slice;
     import mir.random;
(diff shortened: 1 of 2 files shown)


source/mir/random/algorithm.d
```diff
@@ -28,7 +28,7 @@ struct RandomField(G, D, T)
     Constructor.
     Stores the pointer to the `gen` engine.
     +/
-    this()(ref G gen, D var) { _gen = &gen; _var = var; }
+    this()(ref G gen, D var) { _gen = (() @trusted => &gen)(); _var = var; }
```
9il: This is not safe. `gen` can be provided is provided by a user.

9il: Oh, let me think more..

source/mir/random/algorithm.d
```diff
@@ -512,3 +540,27 @@ version(mir_random_test) unittest
     sort(a);
     assert(a == iota(10));
 }
+
+//Above unittest cannot be @safe because mir.ndslice.sorting.sort(...) is not @safe
```
9il: https://github.com/libmir/mir-algorithm/issues/105

source/mir/random/algorithm.d
n8sh: @9il Maybe two constructors: a `@safe` one that accepts `G*`, and a `@system` one that accepts `ref G`. Then it's up to the caller to either obtain the pointer `@safe`ly or mark the code where he obtains the pointer `@trusted` if he wants his code to be `@safe`.

9il:
API should be changed =)
1. Current API should accepts _be value_, but only references RNGs (structure pointers or classes).
2. For local scope user should use generic interface like `auto field(alias rng)(...)`.

n8sh:
>For local scope user should use generic interface like auto field(alias rng)(...).

Can you see what I'm doing wrong here?

```
struct TestField(alias gen)
{
    import mir.random.variable: NormalVariable;
    private NormalVariable!double _var;// <-- Can't slice if present.
    double opIndex()(size_t)
    {
        return gen();
    }
}
@safe version(mir_random_test) unittest
{
    import mir.ndslice: slicedField, slice;
    import mir.random;

    auto rng = Random(unpredictableSeed);
    auto f = TestField!(rng)();
    auto m = f.slicedField(5, 3);// construct lazy matrix 5 row x 3 col
    auto s = m.slice;// allocates data of random matrix
}
```

When I try to compile the test I get this error message:

**[...]/mir-algorithm/source/mir/ndslice/allocation.d(89,8): Error: ret[] is not an lvalue**

**source/mir/random/algorithm.d(204,15): Error: template instance mir.ndslice.allocation.slice!(cast(SliceKind)2, [2LU], FieldIterator!(TestField!(rng))) error instantiating**

If I remove `_var` from `TestField` the compile error goes away.

9il:
I found fix , will publish it soon

9il:
`v0.7.0-alpha8`

PR #47, opened 2017-10-30
Add compatibility with Phobos random interface to PCG and xoroshiro128+

n8sh:
This allows the new engines to be used with legacy libraries that expect
Phobos-style random number generators. It turns out that these most
recent additions are the only engines currently in the library that can
conveniently implement `front` without re-working.

### DIFF
--- a/source/mir/random/engine/pcg.d
+++ b/source/mir/random/engine/pcg.d
@@ -223,11 +223,7 @@ public:
         state = cast(Uint)(acc_mult*state + acc_plus);
     }
 
-    static if (!output_previous)
-    {
-        //InputRange mode is only low-cost when output_previous == true.
-    }
-    else
+    static if (output_previous)
     {
         /++
         Compatibility with Phobos library methods. Presents this RNG as an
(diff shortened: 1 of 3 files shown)


9il:
Great! But can this be generic?

9il:
I mean a wrapper that accepts Mir RNG and propose Phobos RNG?

n8sh:
Sure, could use existing `mir.random.algorithm.RandomRange` and add just a few things to make it a Phobos random.

9il:
#45 should go first then

n8sh:
I think instead doing it at construction would be better, no need to worry about pointers and safety then.

9il:
Generic - LGTM,
Why need specializations for PCG and Xorshift then?

source/mir/random/package.d
```diff
@@ -445,3 +445,105 @@ T randExponential2(T, G)(ref G gen)
...
+{
+    import std.random: isUniformRNG;
+
+    static if (isUniformRNG!Engine)
+    {
+        alias PhobosRandom = Engine;
+    }
+    else
+    {
+        struct PhobosRandom
```
9il: It should be documented thought. (should be also after static if, not else)

9il:
> Why need specializations for PCG and Xorshift then?

Ah, I see why. OK, please move the part that should be documented after static if, not else.

source/mir/random/engine/pcg.d
```diff
@@ -222,6 +222,65 @@ public:
         }
         state = cast(Uint)(acc_mult*state + acc_plus);
     }
+
+    static if (!output_previous)
```
9il: move code so `static if (output_previous) `

n8sh: It looks like ddox doesn't generate documentation for it regardess of where I put it. Irksome.

9il: We use DDOC for docs

source/mir/random/package.d
```diff
@@ -445,3 +445,105 @@ T randExponential2(T, G)(ref G gen)
     auto gen = Xorshift(cast(uint)unpredictableSeed);
     auto v = gen.randExponential2!double();
 }
+
+import std.random: isPhobosUniformRNG = isUniformRNG;
```
9il: please wrap this into template to do not force import std.random files.
```
template isPhobosUniformRNG(T)
{
    import std.random: isUniformRNG;
    alias isPhobosUniformRNG = isUniformRNG!T;
}
```

9il:
Awesome! Thanks!

Issue #49, opened 2017-10-31
mir-random does not compile with latest ldc on raspberry pi


9il:
original https://github.com/libmir/mir/issues/390

9il:
@gizmomogwai could you please specify OS/kernel version?

gizmomogwai:
kernel:
> uname -a
Linux wohnzimmer 4.9.29-10-osmc #1 SMP PREEMPT Mon Oct 9 02:18:18 UTC 2017 armv7l GNU/Linux

this is osmc distribution for raspberry pi 2 (running on rpi 3).

n8sh:
The problem is this:

```
version(linux)
{
//...
    version (X86_64)
        private enum GETRANDOM = 318;
    version (X86)
        private enum GETRANDOM = 355;
//...
```

Since `GETRANDOM` is not defined for other architectures, the outer condition should check architecture instead of just OS.

PR #51, opened 2017-11-01
Fix Linux non-x86/x86_64 compilation failure

n8sh:
https://github.com/libmir/mir-random/issues/49
https://github.com/libmir/mir/issues/390

### DIFF
--- a/source/mir/random/engine/package.d
+++ b/source/mir/random/engine/package.d
@@ -23,6 +23,11 @@ version (OpenBSD)
 version (NetBSD)
     version = GOOD_ARC4RANDOM_BUF;//ChaCha20
 
+version(X86)
+    version = X86_Any;
+version(X86_64)
+    version = X86_Any;
+
 import std.traits;
 
 import mir.random.engine.mersenne_twister;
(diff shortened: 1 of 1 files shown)


9il:
Linux kernel headers define `__NR_getrandom` (and other syscall numbers). I have created an empty https://github.com/libmir/mir-linux-kernel for linux headers. 

https://github.com/torvalds/linux can be used as reference.
We need an `arch/<arch-name->/include/uapi/asm/unistd.h` files and few small others that are imported by some archs.

D file name structure can looks like `mir/linux/arch/<arch-name->/uapi/asm/unistd.d` plus unified file `mir/linux/asm/unistd.d` that predifined arch version from [predefined-versions](https://dlang.org/spec/version.html#predefined-versions) . D does not support all architectures, but many of them are supported by LDC and D specification, see predefined-versions for details.

9il:
If you do not have a time we can merge this for now.

9il:
btw, ping @wilzbach

n8sh:
>If you do not have a time we can merge this for now.

I think adding the syscall numbers for the other architectures is the right thing to do, but I think for now we should unbreak the build.

Issue #41, opened 2017-10-25
Thread local default RNG


9il:
We have `version(D_betterC)` now. This allows to declare a thread local default RNGs like they are in Phobos.

PR #53, opened 2017-11-03
Add thread-local randoms except when D_betterC

n8sh:
For issue https://github.com/libmir/mir-random/issues/41

**(EDIT: below info is outdated. New names are `rne`, `threadLocal!Engine`, `threadLocalPtr!Engine`, and `threadLocalInitialized!Engine`).**
Includes
```
rndGen
threadLocal!Engine
_rawThreadLocal!Engine
```
but I'm not sure it's really desirable to have all of those. `_rawThreadLocal` exposes the static field just so the compiler can tell that it's safe to take the address, if for some reason you need the address of the engine. If there's a way to make the compiler not complain `"cannot take address of ref return of threadLocal() in @safe function"` that would be better.

Also made the documentation look a bit nicer while I was making changes.

### DIFF
--- a/source/mir/random/engine/package.d
+++ b/source/mir/random/engine/package.d
@@ -1,9 +1,49 @@
 /++
+$(SCRIPT inhibitQuickIndex = 1;)
 Uniform random engines.
+
+$(B Sections:)
+        $(LINK2 #Convenience, Convenience)
+&#8226; $(LINK2 #Entropy, Entropy)
+&#8226; $(LINK2 #Traits, Traits)
+&#8226; $(LINK2 #CInterface, C Interface)
+
+$(BOOKTABLE
+
+$(LEADINGROW <a id="Convenience"></a>Convenience)
+$(TR
+    $(RROW Random, Default random number _engine))
+    $(RROW rne, Per-thread uniquely-seeded instance of default `Random`. Requires $(LINK2 https://en.wikipedia.org/wiki/Thread-local_storage, TLS).)
+    $(TR $(TDNW $(LREF threadLocal)`!(Engine)`) $(TD Per-thread uniquely-seeded instance of of any specified `Engine`. Requires $(LINK2 https://en.wikipedia.org/wiki/Thread-local_storage, TLS).))
+
+$(LEADINGROW <a id="Entropy"></a>Entropy)
+$(TR
+    $(RROW unpredictableSeed, Seed of `size_t` using system entropy)
+    $(RROW unpredictableSeedOf, Generalization of `unpredictableSeed` for unsigned integers of different sizes)
+    $(RROW genRandomNonBlocking, Fills a buffer with system entropy, returning number of bytes copied or negative number on error)
+    $(RROW genRandomBlocking, Fills a buffer with system entropy, possibly waiting if the system believes it has insufficient entropy. Returns 0 on success.))
+
+$(LEADINGROW <a id="Traits"></a>Traits)
+$(TR
+    $(RROW isRandomEngine, Check if is random number _engine)
+    $(RROW EngineReturnType, Get return type of random number _engine's `opCall()`)
+    $(RROW isSaturatedRandomEngine, Check if random number _engine `G` such that `G.max == EngineReturnType!(G).max`)
+    $(RROW preferHighBits, Are the high bits of the _engine's output known to have better statistical properties than the low bits?))
+
+$(LEADINGROW <a id="CInterface"></a>C Interface)
+    $(RROW mir_random_engine_ctor, Perform any necessary setup. Automatically called by DRuntime.)
+    $(RROW mir_random_engine_dtor, Release any resources. Automatically called by DRuntime.)
+    $(RROW mir_random_genRandomNonBlocking, External name for $(LREF genRandomNonBlocking))
+    $(RROW mir_random_genRandomBlocking, External name for $(LREF genRandomBlocking))
+)
 
 Copyright: Ilya Yaroshenko 2016-.
 License:  $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
 Authors: Ilya Yaroshenko
+
+Macros:
+    T2=$(TR $(TDNW $(LREF $1)) $(TD $+))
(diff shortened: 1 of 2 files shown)


n8sh:
Oh yeah, I also added `unpredictableSeedOf!T` because I wanted to use it in `threadLocal!Engine` for engines that might accept seeds of larger sizes than size_t.

source/mir/random/engine/package.d
```diff
@@ -199,6 +270,147 @@ version(mir_random_test) unittest
...
+    if you access the generator this way it may not have been seeded yet. But if you need
+    to take the address of the generator, `_rawThreadLocal` is public so the compiler can
+    tell it is `@safe`.
+    +/
+    @property ref Engine threadLocal(Engine)()
+        if (isSaturatedRandomEngine!Engine && is(Engine == struct))
+    {
+        pragma(inline, true);
+        static bool initialized;
+        if (!initialized)
```
9il: This code is not executed in CTFE, right? `_expect` can be helpful here though

source/mir/random/engine/package.d
```diff
@@ -199,6 +270,147 @@ version(mir_random_test) unittest
...
+    used by `threadLocal!Engine`. Normally you will not need to use `_rawThreadLocal`:
+    if you access the generator this way it may not have been seeded yet. But if you need
+    to take the address of the generator, `_rawThreadLocal` is public so the compiler can
+    tell it is `@safe`.
+    +/
+    @property ref Engine threadLocal(Engine)()
+        if (isSaturatedRandomEngine!Engine && is(Engine == struct))
+    {
+        pragma(inline, true);
+        static bool initialized;
```
9il: This variable should be public to do not reproduce the phobos issue. Fibers (green threads) can be interrupted.
So a user code, including mir random functions like `randIndex`, should set `initialized` to false before they start using rng and set it to true after they finish. This will force reinitilization if a thread was terminated during engine state update.

9il: Hm, may be we can do it with a wrapper that hold TLS reference.

source/mir/random/engine/package.d
```diff
@@ -199,6 +270,147 @@ version(mir_random_test) unittest
     static assert(is(EngineReturnType!Random == size_t));
 }
 
+static if (THREAD_LOCAL_STORAGE_AVAILABLE)
+{
+    /++
+    Thread-local instance of the default $(LREF Random) allocated and seeded independently
+    for each thread. Requires $(LINK2 https://en.wikipedia.org/wiki/Thread-local_storage, TLS).
+    +/
+    alias rndGen = threadLocal!Random;
```
9il: How about shorter name without upper case? Maybe `rne`?

9il: `rng` probably is frequently used in user code, but `rne` should not

source/mir/random/engine/package.d
9il: With this change we can promote mir-random as better choice for vibe-d

n8sh: >This variable should be public to do not reproduce the phobos issue. Fibers (green threads) can be interrupted.

Is there a place I can read more about this issue?

>So a user code, including mir random functions like randIndex, should set initialized to false before they start using rng and set it to true after they finish. This will force reinitilization if a thread was terminated during engine state update.

Maybe your "TLS wrapper" idea could make this convenient: the wrapper's constructor (after ensuring the rng is set up) could set `initialized` to false, and the destructor could set `initialized` back to true.

source/mir/random/engine/package.d
n8sh: Renamed it.

n8sh:
Example of what part? What I'm doing might not make sense, because I may have misunderstood the scenario you described. I thought this example was what you had intended:

**(EDIT: Going through step by step, I see that it doesn't actually solve the problem fully.)**

```
function_f () {
auto rng = threadLocal!Random;
//Need a global RNG, so call @property method.
//Get new TL!Random struct. Factory method checks "good state"
//flag shared by every TL!Random on the same thread. If flag is off,
//seeds / reseeds the static random. Then either way, sets "good state"
//flag to false.

int x1 = rng.rand!int;

//Maybe at this point control switches to another thread, and
//maybe because of use of "fibers" it is really still the same thread
//with the same thread-local storage -- I think that's what you were
//describing, right?

//Well, if that happens, and the other thread calls "function_f"
//some other function that uses a TL!Random, when the TL!Random
//is made it will see that "good state" flag is off and so it will
//initialize the static random.

//Now control has come back.
//If another TL!Random did anything to this TL's static random,
//it started by resetting the state ...

// !!!

//...but if the state can get messed up by control switching in the
//middle of a state change, then it's possible that control switched
//back when the other "fiber" was in the middle of such an operation,
//and now we would have the same problem.
// I see now.

int x2 = rng.rand!int;

//Now the TL!Random goes out of scope and its __dtor is called.
//The __dtor sets the "good state" flag to true.
}
```

source/mir/random/engine/package.d
```diff
@@ -199,6 +270,282 @@ version(mir_random_test) unittest
...
+    @nogc nothrow @safe version(mir_random_test) unittest
+    {
+        import mir.random;
+        import std.complex;
+
+        auto c = complex(rne.rand!real, rne.rand!real);
+
+        //If you are going use rne several times in a row,
+        //it's more efficient to retain rne like this:
+        auto rng = rne;
```
9il: Maybe we should prohibit this functionality because is very unsafe? We can do it if we return TL!Engine by reference (its copy constructor already disabled)

9il: If one want best speed hi can init local Engine or use the PTR.

9il:
Looks like I proposed the wrong solution that does not properly fix the issue.

9il:
Sorry about your time was spent on TL. Maybe an public and documented functions that resets the initialised flag should be proposed instead of TL! wrapper. What do you think?

n8sh:
>Maybe an public and documented functions that resets the initialised flag

Yeah I think we need to do this at minimum

PR #55, opened 2017-11-06
Add splitmix64 generator

n8sh:
Aside from the generator itself, it is convenient to have MurmurHash3-style bit mixers predefined.


n8sh:
The reason I use strings instead of enums for `fmix64!"murmurHash3"` etc. was that I find writing `fmix64!FMix64Parameters.murmurHash3` less convenient. Since the template is being resolved at compile-time there isn't a safety advantage to enums here. But if you think using an enum would be better style I can change it.

n8sh:
Using a template with string arguments was unnecessary. Just using aliases instead.

PR #70, opened 2018-02-05
Re-enable passing generators by ref in mir.random.algorithm

n8sh:
The `@safe` way is still there, but now it is an option to do the non-safe way if the programmer prefers the nicer-looking syntax.

Current way, `@safe`:
```d

auto rng = Random(unpredictableSeed);
auto sample = range!rng
    (NormalVariable!double(0, 1))
    .take(10)
    .array;
```

Re-enabled old way, `@system`:
```d
auto rng = Random(unpredictableSeed);
auto sample = rng
    .range(NormalVariable!double(0, 1))
    .take(10)
    .array;
```

The reason for `__traits(compiles, { static assert(isSaturatedRandomEngine!(typeof(gen))); })` in some places instead of `isSaturatedRandomEngine!(typeof(gen))` is to avoid compilation errors.

The `@system` annotations for the alternative overrides are not necessary for the compiler. They're just for the benefit of humans reading the library.

### DIFF
--- a/source/mir/random/algorithm.d
+++ b/source/mir/random/algorithm.d
@@ -13,50 +13,15 @@ import mir.random;
 public import mir.random.engine;
 import mir.random.variable: isRandomVariable;
 
-private enum bool isPassByRef(P) = !is(P == struct) && !is(P == union) && !isPointer!P && !isScalarType!P
-    && (isFunction!P || isDelegate!P || is(P == class) || is(P == interface));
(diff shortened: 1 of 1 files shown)


9il:
Thanks, sorry for breaking changes in the latest release. Feel free to tag new one

9il:
tagged `v0.3.4`

PR #76, opened 2018-02-26
Un-inline fallback code in unpredictableSeed

n8sh:
`unpredictableSeed` includes a block of fallback code which is seldom called. Moving this to a separate function and preventing it from being inlined will improve cache-friendliness.

### DIFF
--- a/source/mir/random/engine/package.d
+++ b/source/mir/random/engine/package.d
@@ -240,6 +240,7 @@ pragma(inline, true)
 @property T unpredictableSeedOf(T)() @trusted nothrow @nogc
     if (isUnsigned!T)
 {
+    import mir.ndslice.internal: _expect;
     T seed = void;
     version (GOOD_ARC4RANDOM_BUF)
     {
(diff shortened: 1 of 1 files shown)


9il:
Lgtm

JohanEngelen:
For this kind of microoptimization, I think it's better to add `likely` and `unlikely` or similar. I've proposed adding a module for these things to Phobos, but haven't worked on it (and can't find the forum thread right now). E.g. something similar to this: https://stackoverflow.com/questions/109710/how-do-the-likely-and-unlikely-macros-in-the-linux-kernel-work-and-what-is-t

n8sh:
We can do both! Those are I think two separate issues, branch prediction and trying to avoid code bloat everywhere `unpredictableSeed` is inlined. There is a wrapper around `llvm_expect` that we use elsewhere and we could use here too:

https://github.com/libmir/mir-algorithm/blob/bc7acc3fc838e4683ba02acb08ca1603b5e9f884/source/mir/ndslice/internal.d#L20-L30
```d
version(LDC)
{
    public import ldc.intrinsics: _expect = llvm_expect;
}
else
{
    T _expect(T)(T val, T expected_val) if (__traits(isIntegral, T))
    {
        return val;
    }
}
```

n8sh:
@JohanEngelen Added `_expect` to the pull request.

n8sh:
Yeah, LDC 1.7.0 will inline it even with `_expect` if `pragma(inline, false)` is removed.

JohanEngelen:
> Yeah, LDC 1.7.0 will inline it even with _expect if pragma(inline, false) is removed.

Thanks for the data point :)

n8sh:
If there aren't objections I'll go ahead and merge this. `unpredictableSeed` is inlined in `threadLocal!Engine`/`rne` so making it slimmer may matter more than one would otherwise assume.

PR #85, opened 2018-05-03
Alter unittest flex `fpEqual` to allow dlang/phobos#6272

n8sh:
For dlang/phobos#6272 which slightly changes result of `std.math.exp`. _Not 100% sure if changing this test is the right thing to do!_

Now on all platforms `fpEqual` requires all but the last two bits of the mantissas to match. Previously (only) with DMD on X86_64 `fpEqual` was exact equality. (The current test does not require exact equality with LDC, but I checked and the test still passes on X86_64 with LDC if `fpEqual` is changed to require exact equality.)

### DIFF
--- a/source/mir/random/flex/internal/area.d
+++ b/source/mir/random/flex/internal/area.d
@@ -21,46 +21,16 @@ FP operations depend on
 */
 version(unittest)
 {
-    // the difference is marginal, but measurable
-    version(Windows)
+    import std.traits : isFloatingPoint;
+    /+
+    Approximate equality. Arguments must match except in the last two bits
+    of the mantissa.
+    +/
+    bool fpEqual(S)(in S a, in S b) @nogc nothrow pure @safe
+    if (isFloatingPoint!S)
     {
-        version = Flex_fpEqual;
-    }
-    else
-    {
-        version(DigitalMars)
-        {
-            version(X86_64)
-            {
-                alias fpEqual = (a, b) => a == b;
-            }
-            else
-            {
-                version = Flex_fpEqual;
-            }
-        }
-        else
-        {
-            version = Flex_fpEqual;
-        }
-    }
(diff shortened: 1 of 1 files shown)


kinke:
Thx @n8sh, much appreciated. Wrt. LDC intrinsic, note that `llvm_exp` simply boils down to a C runtime function call, that's why it may vary from platform to platform: https://godbolt.org/g/iELBRs

https://github.com/libmir/mir-random/blob/master/source/mir/random/flex/package.d#L916 fails too, although it's using `std.math.approxEqual()`. It'd be awesome if you could look into that issue as well.

9il:
@wilzbach please take a look

n8sh:
>https://github.com/libmir/mir-random/blob/master/source/mir/random/flex/package.d#L916 fails too, although it's using `std.math.approxEqual()`.

Very first test with single-precision floats fails with hat area 1.79162e-10 instead of 1.49622e-05. Relaxing the "approxEqual" definition is probably not the right thing to do here. Altering the [arcmean](https://github.com/libmir/mir-random/blob/3c17c14b217f930826d8f475620d995abae91b56/source/mir/random/flex/internal/calc.d#L27) implementation to always use at least double precision trigonometric functions provides enough accuracy for all current tests to pass. Since before we were always using extended precision it's probably okay to do this.

n8sh:
@kinke Addressed the other issue in PR #86

wilzbach:
FWIW the original idea behind this test was too ensure that the Flex algorithm behaves the same on all platform which is very important for this algorithm as otherwise you can end with different intervals and thus different probability function which tends to crash the more "high-level" tests.

OTOH the flex algorithm isn't actively developed right now and as long as the testsuite passes, it shouldn't have a huge impact.
It's still a bit strange that this is necessary and it might affect other people too ...

PR #86, opened 2018-05-04
In arcmean use at least double precision tan/atan (due to dlang/phobos#6272)

n8sh:
Currently `std.math.tan/atan` are calculated using extended precision but dlang/phobos#6272 adds specializations for `float` and `double`. To pass current unittests we need to explicitly use double or extended precision `tan`/`atan` even when `S` is a single-precision.

### DIFF
--- a/source/mir/random/flex/internal/calc.d
+++ b/source/mir/random/flex/internal/calc.d
@@ -27,14 +27,19 @@ References:
 auto arcmean(S)(in ref Interval!S iv)
 {
     import std.math: atan, tan;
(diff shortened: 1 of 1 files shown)


kinke:
Thx guys. @n8sh: Did you happen to run all mir-random tests successfully with that patched Phobos? By the looks of it, we'd need a new 0.46 mir-random release to make the project-tester happy.

n8sh:
@kinke Yes I did, although only on my development machine.

PR #88, opened 2018-05-11
Microsoft Windows DWORD is uint not ulong

n8sh:
https://msdn.microsoft.com/en-us/library/cc230318.aspx

My guess is that when `ULONG_PTR` had to be redefined as `ulong` instead of `uint` the same was done to `DWORD` by accident.


9il:
Sorry, forgot about the PR

PR #90, opened 2018-06-13
Thread-local storage can be available in betterC

n8sh:


9il:
Could you please provide more information here. It is very interesting if this is already true

n8sh:
[This forum post by Walter Bright from June 2017](https://forum.dlang.org/post/oic1et$1jdo$1@digitalmars.com):
>>Does the TLS implementation depend on druntime on any other platforms?
>
>For a C implementation that doesn't support TLS, using it in D with -betterC won't work.

[Followed by this post from Walter Bright](https://forum.dlang.org/post/oid768$qvc$1@digitalmars.com):
>>I'm thinking more of a C implementation where it *does* work. But perhaps you're not expected to do anything besides what you can do in C when it comes to TLS.
>
>It does work with C on Windows, Linux, OSX, and FreeBSD, and so it works with -betterC, too.

----

This little program compiles with `dmd -betterC`/`ldc2 -betterC` and exits with `0`.
```d
int foo()
{
    static int x;
    return ++x;
}

extern(C) int main(int argc, char** argv)
{
    assert(1 == foo());
    return 2 - foo();
}
```

jmh530:
@n8sh What about something like below, but just adding in whatever is known to not have TLS (I have not verified these, just throwing it out there)?

```
version (D_betterC) {
    version(CRuntime_DigitalMars) {
        version = D_betterC_CRuntime_withoutTLS;
    } else version(CRuntime_Glibc) {
        version = D_betterC_CRuntime_withoutTLS;
    } else version(CRuntime_Microsoft) {
        version = D_betterC_CRuntime_withoutTLS;
    }
}

version (D_betterC_CRuntime_withoutTLS) {
    private enum bool THREAD_LOCAL_STORAGE_AVAILABLE = false;	
} else {
    private enum bool THREAD_LOCAL_STORAGE_AVAILABLE = __traits(compiles, { static size_t x = 0; });
}
```

9il:
@n8sh please follow @jmh530 suggestion

n8sh:
@9il Sounds like a plan. I'll perform investigation of specific compiler / C runtime combinations.

PR #91, opened 2018-07-19
In fallBackseed, on x86-64 use RDTSC for ticks, on Alpha use RPCC.

n8sh:
In `fallBackseed`, on x86-64 use RDTSC for ticks, on Alpha use `ldc.intrinsics.readcyclecounter`. Also remove unnecessary failure check for `clock_gettime`.

### DIFF
--- a/source/mir/random/engine/package.d
+++ b/source/mir/random/engine/package.d
@@ -259,32 +259,71 @@ pragma(inline, true)
     return seed;
 }
 
+// Is llvm_readcyclecounter supported on this platform?
+// We need to whitelist platforms where it is known to work because if it
+// isn't supported it will compile but always return 0.
+// https://llvm.org/docs/LangRef.html#llvm-readcyclecounter-intrinsic
+version(LDC)
+{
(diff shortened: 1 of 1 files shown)


source/mir/random/engine/package.d
```diff
@@ -259,32 +259,71 @@ pragma(inline, true)
...
+    }
+    //else version(D_InlineAsm_X86)
+    //{
+    //    // We don't use `rdtsc` with version(D_InlineAsm_X86) because
+    //    // some x86 processors don't support `rdtsc` and because on
+    //    // x86 (but not x86-64) Linux `prctl` can disable a process's
+    //    // ability to use `rdtsc`.
+    //    static assert(0);
+    //}
+    else version(LLVMReadCycleCounter)
```
9il: Any reason to not prefer LLVM first instead of `D_InlineAsm_X86_64`?

n8sh: Thinking about it, no not really. I'll move it first.

9il:
Thanks!

PR #95, opened 2018-09-10
Don't run meson build with Buildkite

Geod24:
```
Buildkite is used by DMD to ensure downstream projects are not broken.
Using meson here forces Buildkite agents to depend on meson.
While it would be possible to do so, that requires some work and coordination,
and might get delayed. In the meantime, just disable meson when buildkit is defined.
```

If possible, I'd like to get this in v1.0.1, as we check out the latest release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ script:
  - dub test --arch "$ARCH"
  - dub build --root=examples/flex_plot
  # - dub build --root=examples/flex_plot/flex_common_pack
- - meson build && cd build && ninja -j4 && ninja test -v && cd ..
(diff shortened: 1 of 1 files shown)


9il:
Fails in travis:
```
$ if [ ${BUILDKITE} != "true" ]; then meson build && cd build && ninja -j4 && ninja test -v && cd ..; done
/home/travis/.travis/job_stages: eval: line 78: syntax error near unexpected token `done'
/home/travis/.travis/job_stages: eval: line 78: `if [ ${BUILDKITE} != "true" ]; then meson build && cd build && ninja -j4 && ninja test -v && cd ..; done '
```

Geod24:
Fixed

PR #97, opened 2018-09-25
SplitMixEngine with constant increment supports `shared`

n8sh:


9il:
LGTM, can be merged and tagged

Issue #99, opened 2018-09-27
Statistical artifact


luismarques:
Hi. I'm getting a statistical artifact in a simulation of a quantile estimator. This artifact appears when I use mir-random, but not when I use Phobos.

At first I thought it was just an artifact (more or less expected) of an imperfect pseudo-random number generator. So I expected that if I changed the generator, either the artifact would go away or it would be replaced by different artifacts. But I tried two different mir generators (pcg32 and SplitMixEngine) and the artifact remains. Since this doesn't happen with Phobos, it suggests it might be an issue with mir-random.

Here's my simulation procedure:

1. Generate a random sample of size `n` (`n=30` in my examples), using some distribution (at first I was using an `ExponentialVariable`, but I changed it to a `UniformVariable`, so I could easily compare it with Phobos).

2. Generate a random quantile between 0 and 1.

3. Estimate the quantile value using the sample. To do that, we sort the sample (to obtain the order statistics) and select the ordered sample value that best estimates the sample quantile (basically, the kth order statistic estimates `q=k/(n+1)`).

4. Since we know the actual distribution, obtain the true quantile corresponding to the estimated quantile value.

5. In one file (`quantile.txt`) output the pair (`requested quantile`, `estimated quantile`)

6. In another file (`quantile-diff.txt`) output the pair (`requested quantile`, `estimated quantile` - `requested quantile`)

7. Plot both files, using gnuplot

Here's an example of the kind of artifact I mean, a visible line among the expected dot dispersion. You might need to run the simulation more than once for the artifact to show, or for it to show in a different region.

Requested vs estimated quantile:
![quantile-1](https://user-images.githubusercontent.com/171812/46144853-1af46380-c256-11e8-99ed-7cabc1989d7f.png)

Quantile diff (same simulation run):
![quantile-diff-1](https://user-images.githubusercontent.com/171812/46144980-81798180-c256-11e8-81ee-9354c7dc3335.png)

As part of my efforts to isolate this issue, one of the things I tried was to use a fixed order statistic (instead of the one that best approximates each quantile). Here's some example data I obtained using the order statistic x_{3:30} (D array index 2).

One trial without the artifact:

![quantile-2](https://user-images.githubusercontent.com/171812/46145369-e08bc600-c257-11e8-9071-547d5e95ffba.png)
![quantile-diff-2](https://user-images.githubusercontent.com/171812/46145374-e5507a00-c257-11e8-89a0-94b86d4aa88e.png)

One trial with the artifact:
[...]

PR #100, opened 2018-09-29
fix issue #99

9il:
- Rework API for `randomSlice` (full rework), `shuffle` (accepts only Slice), `sample` (fix param order).
 - Remove `range` and `field` because they can be either secure or consistent with Phobos/Mir.
 - Update examples in the readme.

### DIFF
--- a/source/mir/random/algorithm.d
+++ b/source/mir/random/algorithm.d
@@ -5,897 +5,243 @@ License:  $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
 +/
 module mir.random.algorithm;
 
-import std.range.primitives;
-import std.traits;
 import mir.math.common;
-
+import mir.ndslice.slice;
+import mir.primitives: hasLength;
 import mir.random;
-public import mir.random.engine;
-import mir.random.variable: isRandomVariable;
 import mir.random.ndvariable: isNdRandomVariable;
-
-// Removed documentation because no longer needed
-// but leaving in code in case anyone was using it.
-public template isReferenceToSaturatedRandomEngine(G)
-{
-    static if (isPointer!G)
-        enum bool isReferenceToSaturatedRandomEngine = isSaturatedRandomEngine!(typeof(((G g) => *g)(G.init)));
-    else static if (is(G == class) || is(G == interface))
-        enum bool isReferenceToSaturatedRandomEngine = isSaturatedRandomEngine!G;
-    else
-        enum bool isReferenceToSaturatedRandomEngine = false;
-}
+import mir.random.variable: isRandomVariable;
+import std.range.primitives: isInputRange, isForwardRange;
+import std.traits;
+public import mir.random.engine;
 
 /++
 Allocates ndslice (vector, matrix, or tensor) and fills it with random numbers.
+If no variable is specified each element `e` is generated per `rand!(typeof(e))`.
 
 Params:
     gen = random engine (optional, param or template param)
     var = random variable (optional)
     lengths = one or more lengths
 +/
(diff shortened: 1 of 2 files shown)


9il:
ping @n8sh

n8sh:
>`shuffle` (accepts only Slice)

What's the motivation for this change?

source/mir/random/algorithm.d
```diff
@@ -34,868 +23,224 @@ Params:
     var = random variable (optional)
     lengths = one or more lengths
 +/
-auto randomSlice(T, G, D, size_t N)(G gen, D var, size_t[N] lengths...)
-    if (isSaturatedRandomEngine!G && isRandomVariable!D &&
+pragma(inline, false)
```
n8sh: Why `pragma(inline, false)`?

9il:
> What's the motivation for this change?

Any random access range with length in Mir can be constructed as Slice. If it is not a Slice it can be converted to Slice. No plans to support Phobos RoR API in general.

source/mir/random/algorithm.d
9il: It will generate quite a large chunk of code that may be frequently used in different places. The pragma helps to reduce code size. Mir based programs should be huge as vibe.d

source/mir/random/algorithm.d
```diff
@@ -34,868 +23,224 @@ Params:
...
+auto randomSlice(G, D, size_t N)(G gen, D var, size_t[N] lengths...)
+    if (N && isSaturatedRandomEngine!G && isRandomVariable!D &&
         (is(G == class) || is(G == interface)))
 {
-    import mir.ndslice.allocation: slice;
-    import mir.ndslice.slice: slicedField;
-    return field(gen, var).slicedField(lengths).slice;
+    import mir.ndslice.allocation: uninitSlice;
+    alias T = typeof(var(gen));
+    auto ret = (()@trusted => lengths.uninitSlice!T)();
```
n8sh: You don't need a `@trusted` lambda here. `uninitSlice` uses `std.array.uninitializedArray` which is `@trusted` if `T` has no indirections.

n8sh: Same goes for the rest below. I can remove them directly though.

9il: Thanks!

source/mir/random/algorithm.d
```diff
@@ -34,868 +23,224 @@ Params:
...
 {
-    import mir.ndslice.allocation: slice;
-    import mir.ndslice.slice: slicedField;
-    return field(gen, var).slicedField(lengths).slice;
+    import mir.algorithm.iteration: each;
+    import mir.ndslice.allocation: uninitSlice;
+    import mir.ndslice.topology: pack;
+    alias T = D.Element;
+    auto ret = lengths.uninitSlice!T();
+    ret.pack!1.each!(a => var(gen, a));
```
n8sh: Checks out.

source/mir/random/algorithm.d
```diff
@@ -34,868 +23,224 @@ Params:
...
-    auto sample2 = randomSlice!cdouble(var, 5, 3);
-
-    Random rng = Random(12345);
-    auto sample3 = rng.randomSlice!cdouble(var, 5, 3); //rng safely passed by ref
-}
-
-/// Random binary data
-nothrow @safe version(mir_random_test) unittest
+pragma(inline, false)
+auto randomSlice(T, G, size_t N)(G gen, size_t[N] lengths...)
```
n8sh: It might be worth mentioning in the documentation that for this family of functions, if `T` is floating point each result will be in the interval `(-1.0, 1.0)`.

PR #101, opened 2018-10-03
Fix documentation of rand!IntType: min result is IntType.min not 0

n8sh:
I assume the documentation is wrong, but if it is not then the behavior must be fixed instead.


9il:
thanks!

PR #102, opened 2018-10-23
Split mir.random.engine.xoshiro from mir.random.engine.xorshift

n8sh:
The moved generators are publicly imported by `mir.random.engine.xorshift` so existing programs expecting them there will still work.


9il:
LGTM, can be merged

PR #107, opened 2018-11-08
Optimize rand!enum when using 32-bit RNG and/or # of elements is power-of-2

n8sh:


source/mir/random/package.d
```diff
@@ -199,14 +219,35 @@ T rand(T, G)(scope ref G gen)
         enum tiny = [EnumMembers!T] == [Iota!(EnumMembers!T.length)];
     else
         enum tiny = false;
+    enum n = [EnumMembers!T].length;
+    // If `gen` produces 32 bits or fewer at a time and we have fewer
+    // than 2^^32 elements, use a `uint` index.
+    static if (n <= uint.max && EngineReturnType!G.max <= uint.max && uint.max <= size_t.max)
```
9il: The last condition is always true

n8sh: I considered maybe GDC might support 16-bit platforms

n8sh: even though by the specification size_t is always supposed to be at least uint

PR #111, opened 2018-11-14
Rename unpredictableSeedOf!UIntType to unpredictableSeed!UIntType

n8sh:
Fewer letters and matches new overload in Phobos std.random (https://github.com/dlang/phobos/pull/6580).


9il:
needs `deprecated`  alias

n8sh:
Added `deprecated` to the alias.

PR #113, opened 2018-11-14
Lazily detect Linux getrandom support instead of explicit version check at startup

n8sh:
Assume getrandom is available to start. If getrandom fails with `errno == ENOSYS` switch to the /dev/random implementation and don't try getrandom again. This approach simplifies the code and performs less work in the typical case where the Linux version >= 3.17.

### DIFF
--- a/source/mir/random/engine/package.d
+++ b/source/mir/random/engine/package.d
@@ -674,46 +674,8 @@ else
 
 static if (LINUX_NR_GETRANDOM)
 {
-    private enum GET_RANDOM {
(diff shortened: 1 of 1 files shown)


9il:
does it break purity?

n8sh:
It does not because the function was already impure.
* The code was previously using `__gshared hasGetRandom` which was set in a static ctor.
* The GETRANDOM syscall itself is impure.

9il:
OK, LGTM then. Please merge when ready

Issue #117, opened 2019-09-29
multinomial


fccoelho:
any plans to include multinomial random variable?

9il:
We don't, but we can review and accept a pull request with this random varialble.

jmh530:
This shouldn't be that hard to implement on your own given `BinomialVariable`. According to the r function's documentation [1], they calculate it as:

> The rmultinom() algorithm draws binomials X[j] from Bin(n[j], P[j]) sequentially, where n[1] = N (N := size), P[1] = p[1] (p is prob scaled to sum 1), and for j ≥ 2, recursively, n[j] = N - sum(k=1, …, j-1) X[k] and P[j] = p[j] / (1 - sum(p[1:(j-1)])). 

[1] https://stat.ethz.ch/R-manual/R-devel/library/stats/html/Multinom.html

fccoelho:
Ok, 

I'll be glad to contribute. I have started an implementation ported straight from [GSL](https://github.com/ampl/gsl/blob/master/randist/multinomial.c).
the code is here: https://gist.github.com/run-dlang/edc536446341a957256d1bfeeefaf1da
However, there is still some bug in it, if someone could have a look I'd **greatly** appreciate it.
After we get this working, It'll still need to be modified to conform with `mir.random`'s API, but I believe that as is it is easier to test and make sure is working properly.

jmh530:
It's like you port it more-or-less straight from GSL. The only line I'm
concerned about is
auto rng = Random(25923);
I would think you would instead have something like
alias rng = rne;
but I don't use mir.random that much as of now...so I don't know what is
the recommended approach.

As you say, it would definitely need to be modified to be part of
mir.random's API.

jmh530:
One other change it needs:
change
[1.0/6, 2.0/6, 3/6]
to
[1.0/6, 2.0/6, 3.0/6]

fccoelho:
Thanks @jmh530, for spotting the errors! It seems to be working fine now, I'll come up with some unit tests now to make sure it is probabilistically correct, i.e., the sampled values match theoretical expectations.

Meanwhile, should I use [this](https://github.com/libmir/mir-random/blob/master/source/mir/random/variable.d} as a code template, for the pull request? 
I am new to D so I'll have to take some time to digest some of the specifics of the mir-random code.

jmh530:
[...]

PR #118, opened 2019-10-05
Implementation of Multinomial random variable

fccoelho:
I have ported this from GSL as described on issue #117.
it has been adapted to Mir-random API and has a simple unittest which are passing.

There is a minor deviation from the way other ndvariables in that the `opCall` is not void, but rather returns the result expected. I try but could not make it alter the result variable passed in in the outer scope, even though my implementation follows exactly the other variable's implementations. The only difference is that multinomial returns an array of long integers instead of doubles.
If there's a way to fix this, please let me know and I'll change the code, but as it is  the function works correctly. It is worth pointing out that the `opCall` of the scalar random variables, defined in `variable.d`, are also not void, returning the results directly.

### DIFF
--- a/source/mir/random/ndvariable.d
+++ b/source/mir/random/ndvariable.d
@@ -283,7 +283,7 @@ nothrow @safe version(mir_random_test) unittest
 /++
 Multinomial distribution.
 +/
-struct MultinomialVariable(U:size_t, T)
+struct MultinomialVariable(T)
     if (isFloatingPoint!T)
 {
     import mir.random.variable : binomialVar;
@@ -296,6 +296,8 @@ struct MultinomialVariable(U:size_t, T)
     ///
     const(T)[] probs;
     size_t N;
+    double norm = 0.0;
+
 
     /++
     Params:
@@ -307,30 +309,30 @@ struct MultinomialVariable(U:size_t, T)
     {
         this.N = N;
         this.probs = probs;
+         /// Makes sure probabilities add up to one, by calculating a normalization factor
+        foreach(k, p; this.probs)
+        {
+            this.norm += p;
+        }
+        //assert(fabs(this.norm - 1) <= double.min_normal * this.probs.length * 2);
     }
 
     ///
     pragma(inline, false)
-    void opCall(G)(scope ref G gen, scope U[] result)
+    void opCall(G)(scope ref G gen, scope size_t[] result)
         if (isSaturatedRandomEngine!G)
     {
-        double norm = 0.0;
         double sum_p = 0.0;
 
         uint sum_n = 0;
 
 
-        /// Makes sure probabilities add up to one, by calculating a normalization term
-        foreach(k, p; this.probs)
-        {
-            norm += p;
-        }
 
         foreach(k, p; this.probs)
         {
             if (p > 0.0)
             {
-                auto rv = binomialVar!T(this.N - sum_n, p / (norm - sum_p));
+                auto rv = binomialVar!T(this.N - sum_n, p / (this.norm - sum_p));
                 result[k] = cast(uint)rv(gen);
 
             }
@@ -346,7 +348,7 @@ struct MultinomialVariable(U:size_t, T)
 
     }
     /// ditto
-    void opCall(G)(scope G* gen, scope U[] result)
+    void opCall(G)(scope G* gen, scope size_t[] result)
         if (isSaturatedRandomEngine!G)
     {
         pragma(inline, true);
@@ -355,7 +357,7 @@ struct MultinomialVariable(U:size_t, T)
 }
 
 /// ditto
-MultinomialVariable!(U,T) multinomialVar(U, T)(in U N, in T[] probs)
+MultinomialVariable!(T) multinomialVar(T)(size_t N, return const T[] probs)
     if (isFloatingPoint!T)
 {
     return typeof(return)(N, probs);
@@ -370,7 +372,7 @@ alias multinomialVariable = multinomialVar;
 nothrow @safe version(mir_random_test) unittest
 {
     size_t s = 10000;
-    double[6] p =[1/6., 1/6., 1/6, 1/6., 1/6., 1/6.];
+    double[6] p =[1/6., 1/6., 1/6, 1/6., 1/6., 1/16.]; // probs add to less than one
     auto rv = multinomialVar(s, p);
     size_t[6] x;
     rv(rne,x[]);
@@ -382,7 +384,7 @@ nothrow @safe version(mir_random_test) unittest
 {
     Random* gen = threadLocalPtr!Random;
     size_t s = 1000;
-    auto rv = MultinomialVariable!(size_t, double)(s, [1.0, 5.7, 0.3]);
+    auto rv = MultinomialVariable!(double)(s, [1.0, 5.7, 0.3]); // probs add to more than one
     size_t[3] x;
     rv(gen,x[]);
     assert(x[0]+x[1]+x[2] == s);
--- a/source/mir/random/ndvariable.d
+++ b/source/mir/random/ndvariable.d
@@ -283,7 +283,7 @@ nothrow @safe version(mir_random_test) unittest
 /++
 Multinomial distribution.
 +/
-struct MultinomialVariable(U:ulong, T)
+struct MultinomialVariable(U:size_t, T)
     if (isFloatingPoint!T)
 {
     import mir.random.variable : binomialVar;
@@ -295,7 +295,7 @@ struct MultinomialVariable(U:ulong, T)
 
     ///
     const(T)[] probs;
-    const ulong N;
+    const size_t N;
 
     /++
     Params:
@@ -303,9 +303,7 @@ struct MultinomialVariable(U:ulong, T)
         N = Number of rolls
     Constraints: `sum(probs[i]) <= 1`
     +/
-
-    /// ditto
-    this()(const ulong N, const(T)[] probs)
+    this()(const size_t N, const(T)[] probs)
     {
         this.N = N;
         this.probs = probs;
@@ -313,7 +311,7 @@ struct MultinomialVariable(U:ulong, T)
 
     ///
     pragma(inline, false)
-    ulong[] opCall(G)(scope ref G gen, scope U[] result)
+    size_t[] opCall(G)(scope ref G gen, scope U[] result)
         if (isSaturatedRandomEngine!G)
     {
         uint k;
@@ -324,7 +322,7 @@ struct MultinomialVariable(U:ulong, T)
         alias rng = rne;
 
         uint sum_n = 0;
-        ulong[] n;
+        size_t[] n;
 
         /* p[k] may contain non-negative weights that do not sum to 1.0.
        * Even a probability distribution will not exactly sum to 1.0
@@ -341,6 +339,7 @@ struct MultinomialVariable(U:ulong, T)
         {
             if (p[k] > 0.0)
             {
+
                 auto rv = binomialVar(this.N - sum_n, p[k] / (norm - sum_p));
                 n[k] = rv(rng);
 
@@ -358,7 +357,7 @@ struct MultinomialVariable(U:ulong, T)
 
     }
     /// ditto
-    ulong[] opCall(G)(scope G* gen, scope U[] result)
+    size_t[] opCall(G)(scope G* gen, scope U[] result)
         if (isSaturatedRandomEngine!G)
     {
         pragma(inline, true);
@@ -381,11 +380,10 @@ alias multinomialVariable = multinomialVar;
 ///
 nothrow @safe version(mir_random_test) unittest
 {
-
-    ulong s = 1000;
+    size_t s = 1000;
     double[3] p =[1.0, 5.7, 0.3];
     auto rv = multinomialVar(s, p);
-    ulong[3] x;
+    size_t[3] x;
     x = rv(rne,x[]);
     //writeln(x);
     assert(x[0]+x[1]+x[2] == s);
@@ -395,9 +393,9 @@ nothrow @safe version(mir_random_test) unittest
 nothrow @safe version(mir_random_test) unittest
 {
     Random* gen = threadLocalPtr!Random;
-    ulong s = 1000;
-    auto rv = MultinomialVariable!(ulong,double)(s, [1.0, 5.7, 0.3]);
-    ulong[3] x;
+    size_t s = 1000;
+    auto rv = MultinomialVariable!(size_t,double)(s, [1.0, 5.7, 0.3]);
+    size_t[3] x;
     x = rv(gen,x[]);
     assert(x[0]+x[1]+x[2] == s);
 }
--- a/source/mir/random/ndvariable.d
+++ b/source/mir/random/ndvariable.d
@@ -299,7 +299,7 @@ struct MultinomialVariable(U:size_t, T)
 
     /++
     Params:
-        probs = probabilities
+        probs = probabilities of the multinomial distribution
         N = Number of rolls
     Constraints: `sum(probs[i]) <= 1`
     +/
@@ -320,17 +320,13 @@ struct MultinomialVariable(U:size_t, T)
         alias rng = rne;
 
         uint sum_n = 0;
-        //size_t[] n;
 
-        /* p[k] may contain non-negative weights that do not sum to 1.0.
-       * Even a probability distribution will not exactly sum to 1.0
-       * due to rounding errors.
-       */
 
+        /// Makes sure probabilities add up to one, by calculating a normalization term
         foreach(k, p; this.probs)
         {
 
-            norm += p;//this.probs[k];
+            norm += p;
         }
 
         foreach(k, p; this.probs)
@@ -350,8 +346,7 @@ struct MultinomialVariable(U:size_t, T)
             sum_p += p;
             sum_n += result[k];
         }
-        //result = n;
-        //return result;
+
 
     }
     /// ditto
@@ -383,7 +378,6 @@ nothrow @safe version(mir_random_test) unittest
     auto rv = multinomialVar(s, p);
     size_t[3] x;
     rv(rne,x[]);
-    //writeln(x);
     assert(x[0]+x[1]+x[2] == s);
 }
 
--- a/source/mir/random/ndvariable.d
+++ b/source/mir/random/ndvariable.d
@@ -296,7 +296,6 @@ struct MultinomialVariable(T)
     ///
     const(T)[] probs;
     size_t N;
-    double norm = 0.0;
 
 
     /++
@@ -334,7 +333,7 @@ struct MultinomialVariable(T)
         {
             if (p > 0.0)
             {
-                auto rv = binomialVar!T(this.N - sum_n, p / (this.norm - sum_p));
+                auto rv = binomialVar!T(this.N - sum_n, p / (1 - sum_p));
                 result[k] = cast(uint)rv(gen);
 
             }
--- a/source/mir/random/ndvariable.d
+++ b/source/mir/random/ndvariable.d
@@ -314,10 +314,8 @@ struct MultinomialVariable(U:size_t, T)
     void opCall(G)(scope ref G gen, scope U[] result)
         if (isSaturatedRandomEngine!G)
     {
-        auto K = this.probs.length;
         double norm = 0.0;
         double sum_p = 0.0;
-        alias rng = rne;
 
         uint sum_n = 0;
 
@@ -325,7 +323,6 @@ struct MultinomialVariable(U:size_t, T)
         /// Makes sure probabilities add up to one, by calculating a normalization term
         foreach(k, p; this.probs)
         {
-
             norm += p;
         }
 
@@ -333,9 +330,8 @@ struct MultinomialVariable(U:size_t, T)
         {
             if (p > 0.0)
             {
-
                 auto rv = binomialVar!T(this.N - sum_n, p / (norm - sum_p));
-                result[k] = cast(uint)rv(rng);
+                result[k] = cast(uint)rv(gen);
 
             }
             else
@@ -373,12 +369,12 @@ alias multinomialVariable = multinomialVar;
 ///
 nothrow @safe version(mir_random_test) unittest
 {
-    size_t s = 1000;
-    double[3] p =[1.0, 5.7, 0.3];
+    size_t s = 10000;
+    double[6] p =[1/6., 1/6., 1/6, 1/6., 1/6., 1/6.];
     auto rv = multinomialVar(s, p);
-    size_t[3] x;
+    size_t[6] x;
     rv(rne,x[]);
-    assert(x[0]+x[1]+x[2] == s);
+    assert(x[0]+x[1]+x[2]+x[3]+x[4]+x[5] == s);
 }
 
 ///
--- a/source/mir/random/ndvariable.d
+++ b/source/mir/random/ndvariable.d
@@ -291,11 +291,11 @@ struct MultinomialVariable(U:size_t, T)
     ///
     enum isNdRandomVariable = true;
     ///
-    alias Element = T;
+    alias Element = uint;
 
     ///
     const(T)[] probs;
-    const size_t N;
+    size_t N;
 
     /++
     Params:
@@ -303,7 +303,7 @@ struct MultinomialVariable(U:size_t, T)
         N = Number of rolls
     Constraints: `sum(probs[i]) <= 1`
     +/
-    this()(const size_t N, const(T)[] probs)
+    this()(size_t N, const(T)[] probs)
     {
         this.N = N;
         this.probs = probs;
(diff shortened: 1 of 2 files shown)


source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
     assert(fabs(x[0] + x[1] + x[2] - 1) < 1e-10);
 }
 
+/++
+Multinomial distribution.
++/
+struct MultinomialVariable(U:size_t, T)
```
9il: `struct MultinomialVariable(T)`

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+    import mir.random.variable : binomialVar;
+
+    ///
+    enum isNdRandomVariable = true;
+    ///
+    alias Element = T;
+
+    ///
+    const(T)[] probs;
+    const size_t N;
```
9il: `size_t N`

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+
+    /++
+    Params:
+        probs = probabilities
+        N = Number of rolls
+    Constraints: `sum(probs[i]) <= 1`
+    +/
+    this()(const size_t N, const(T)[] probs)
```
9il: `this()(size_t N, const(T)[] probs)`

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+        this.probs = probs;
+    }
+
+    ///
+    pragma(inline, false)
+    size_t[] opCall(G)(scope ref G gen, scope U[] result)
+        if (isSaturatedRandomEngine!G)
+    {
+        uint k;
+        alias p =this.probs;
```
9il: don't use additional aliases

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+        this.N = N;
+        this.probs = probs;
+    }
+
+    ///
+    pragma(inline, false)
+    size_t[] opCall(G)(scope ref G gen, scope U[] result)
+        if (isSaturatedRandomEngine!G)
+    {
+        uint k;
```
9il: no need this.

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+        alias p =this.probs;
+        auto K = p.length;
```
9il: ditto

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+        double norm = 0.0;
+        double sum_p = 0.0;
+        alias rng = rne;
+
+        uint sum_n = 0;
+        size_t[] n;
```
9il: no need this, the `result` should be initialized instead, it should be already preallocated.

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+    +/
+    this()(const size_t N, const(T)[] probs)
+    {
+        this.N = N;
+        this.probs = probs;
+    }
+
+    ///
+    pragma(inline, false)
+    size_t[] opCall(G)(scope ref G gen, scope U[] result)
```
9il: `void opCall(G)(scope ref G gen, scope uint[] result)`

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
++/
+struct MultinomialVariable(U:size_t, T)
+    if (isFloatingPoint!T)
+{
+    import mir.random.variable : binomialVar;
+
+    ///
+    enum isNdRandomVariable = true;
+    ///
+    alias Element = T;
```
9il: `alias Element = uint;`

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+    pragma(inline, false)
+    size_t[] opCall(G)(scope ref G gen, scope U[] result)
+        if (isSaturatedRandomEngine!G)
+    {
+        uint k;
+        alias p =this.probs;
+        auto K = p.length;
+        double norm = 0.0;
+        double sum_p = 0.0;
+        alias rng = rne;
```
9il: ditto

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+
+        uint sum_n = 0;
+        size_t[] n;
+
+        /* p[k] may contain non-negative weights that do not sum to 1.0.
+       * Even a probability distribution will not exactly sum to 1.0
+       * due to rounding errors.
+       */
+
+        for (k = 0; k < K; k++)
```
9il: common `foreach` loop should be used instead.

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+        {
+            n ~= 0; /// intializing array n
+            norm += p[k];
+        }
+
+        for (k = 0; k < K; k++)
```
9il: ditto

source/mir/random/ndvariable.d
9il: `foreach(k, p; probs)`

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+        {
+            if (p[k] > 0.0)
+            {
+
+                auto rv = binomialVar(this.N - sum_n, p[k] / (norm - sum_p));
```
9il: `binomialVar!T`

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+                n[k] = rv(rng);
```
9il: `n[k] = cast(uint)rv(rng);`

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,126 @@ nothrow @safe version(mir_random_test) unittest
...
+
+            sum_p += p[k];
+            sum_n += n[k];
+        }
+        result = n;
+        return result;
+
+    }
+    /// ditto
+    size_t[] opCall(G)(scope G* gen, scope U[] result)
```
9il: `void opCall(G)(scope G* gen, scope uint[] result)`

source/mir/random/ndvariable.d
fccoelho: You mean the `if (isSaturatedRandomEngine!G)`?

9il:
GSL likely isn't compatible with BSL-1.0. So, the direct port can't be accepted.
However, if you understand the math and can implement the algorithm from scratch, then it can be accepted.

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

fccoelho:
> GSL likely isn't compatible with BSL-1.0. So, the direct port can't be accepted.
> However, if you understand the math and implement the algorithm from scratch, then it can be accepted.

Regarding this licensing issue, the [GSL implementation](https://github.com/ampl/gsl/blob/master/randist/multinomial.c) comes straight from this reference:
```
 C.S. David, The computer generation of multinomial random variates,
   Comp. Stat. Data Anal. 16 (1993) 205-217
```
The algorithm is so simple that there is no other way I see to implement it efficiently.
I can bet that all other implementations of the RNG are based on the same algorithm.

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,118 @@ nothrow @safe version(mir_random_test) unittest
...
+        this.N = N;
+        this.probs = probs;
+    }
+
+    ///
+    pragma(inline, false)
+    void opCall(G)(scope ref G gen, scope U[] result)
+        if (isSaturatedRandomEngine!G)
+    {
+        auto K = this.probs.length;
```
jmh530: Is this needed?

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,118 @@ nothrow @safe version(mir_random_test) unittest
...
+        double norm = 0.0;
+        double sum_p = 0.0;
+        alias rng = rne;
+
+        uint sum_n = 0;
+
+
+        /// Makes sure probabilities add up to one, by calculating a normalization term
```
jmh530: Why not use `auto norm = sum(this.probs);`? Then you can do something like `typeof(norm) sum_p = 0;`.

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,118 @@ nothrow @safe version(mir_random_test) unittest
...
+
+    ///
+    pragma(inline, false)
+    void opCall(G)(scope ref G gen, scope U[] result)
+        if (isSaturatedRandomEngine!G)
+    {
+        auto K = this.probs.length;
+        double norm = 0.0;
+        double sum_p = 0.0;
+        alias rng = rne;
```
jmh530: I had included `alias rng = rne;` in the other version just to get it working. Not sure needed here.

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,118 @@ nothrow @safe version(mir_random_test) unittest
...
+            norm += p;
+        }
+
+        foreach(k, p; this.probs)
+        {
+            if (p > 0.0)
+            {
+
+                auto rv = binomialVar!T(this.N - sum_n, p / (norm - sum_p));
+                result[k] = cast(uint)rv(rng);
```
jmh530: Should this be `result[k] = cast(uint)rv(gen);`?

source/mir/random/ndvariable.d
fccoelho: to have `sum()` I need to import it from mir-ndslice. is it worth it?

source/mir/random/ndvariable.d
fccoelho: yes, fixed.

source/mir/random/ndvariable.d
fccoelho: removed that

source/mir/random/ndvariable.d
fccoelho: not anymore. removed.

source/mir/random/ndvariable.d
jmh530: I would leave that up to @9il

9il: `sum` is from `mir-algorithm` that is an optional dependency. So, it shouldn't be imported in this module.

source/mir/random/ndvariable.d
9il: It should be strictly `uint`, we shouldn't use `size_t` here at all

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,114 @@ nothrow @safe version(mir_random_test) unittest
...
+        if (isSaturatedRandomEngine!G)
+    {
+        double norm = 0.0;
+        double sum_p = 0.0;
+
+        uint sum_n = 0;
+
+
+        /// Makes sure probabilities add up to one, by calculating a normalization term
+        foreach(k, p; this.probs)
```
9il: Lets move this loop into the constructor and check with `assert(fabs(norm - 1) <= double.min_normal * probs.length * 2)`

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,114 @@ nothrow @safe version(mir_random_test) unittest
...
+    void opCall(G)(scope G* gen, scope U[] result)
+        if (isSaturatedRandomEngine!G)
+    {
+        pragma(inline, true);
+        return opCall(*gen, result);
+    }
+}
+
+/// ditto
+MultinomialVariable!(U,T) multinomialVar(U, T)(in U N, in T[] probs)
```
9il: `MultinomialVariable!(T) multinomialVar(T)(in size_t N, in T[] probs)`

9il:
few nitpicks, other LGRM

source/mir/random/ndvariable.d
fccoelho: We currently don't have a constructor, at least not an explicit one: `this(){...}`
Should we add one?

fccoelho: This assert is not correct. `norm` is the sum of values passed in the probs array. If they don't add to one, the code in the main loop, `p / (norm - sum_p)` will make sure `p`s add to one.  The test in the second unittest is easy to calculate by hand to verify this.

fccoelho: But I have moved the loop into the constructor. It cleans up the main calling function by removing a computation that only need to be done on initialization.

jmh530: @fccoelho I was initially sympathetic to @9il's point here. If norm is different from 1, then there must be some reason for that. I would rather the user be alerted to the issue and solve the problem on their own. However, I also noticed that the code itself does not have any restriction that the p's are less than 1. I don't see anything that would prevent passing something like p = [10, 30, 60] to Multinomial. I'm not sure if this is intended or not.

fccoelho: Well, it seems that some protection against probs that don't add up to one is necessary that's what the norm term is used for. One motivation I can think of for being able to parameterize with arbitrary numbers (which don't add up to one), would be to simplify creating an equivalent multinomial generator from a sample generated by another one, of which you don't necessarily know the probs.

9il: From the API point of view the probs should be already normalized. Otherwise it is weights. Mir is positioned as a library with low level abilities and maximal control for a user to create his own wrappers on top of that. Probs are preferred.

fccoelho: We can simply assert probs sum up to 1, but then we expose the call to a runtime error if probs are not carefully specified. The way it is implemented right now, it will never throw such an error. So it would be safer without any extra computational cost, since for the assert we would have to sum the probs anyway. Besides, there would be no impact to users who wish to pass exact probabilities.

source/mir/random/ndvariable.d
fccoelho: Done

source/mir/random/ndvariable.d
fccoelho: Done

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,116 @@ nothrow @safe version(mir_random_test) unittest
...
+        //assert(fabs(this.norm - 1) <= double.min_normal * this.probs.length * 2);
+    }
+
+    ///
+    pragma(inline, false)
+    void opCall(G)(scope ref G gen, scope size_t[] result)
+        if (isSaturatedRandomEngine!G)
+    {
+        double sum_p = 0.0;
+
```
9il: no need empty line here

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,116 @@ nothrow @safe version(mir_random_test) unittest
...
+        uint sum_n = 0;
+
+
+
```
9il: ditt

source/mir/random/ndvariable.d
9il: Just place the loop and assert itself into `version(assert)` block
```
version(assert)
{
 /// foreach loop here
}
```

fccoelho: Ok, but how do I specify in the unittests which version should be tested?
By the way, is this the way it should be done?
```D
this()(size_t N, const(T)[] probs)
    {
        this.N = N;
        this.probs = probs;
         /// Makes sure probabilities add up to one, by calculating a normalization factor
        version(assert)
        {
            foreach(k, p; this.probs)
        {
            this.norm += p;
        }
            assert(fabs(norm - 1) <= double.min_normal * probs.length * 2);
        }
        else
        {
            foreach (k, p; this.probs)
            {
                this.norm += p;
            }
        }
        //assert(fabs(this.norm - 1) <= double.min_normal * this.probs.length * 2);
    }
```

9il:
```d
    this()(size_t N, const(T)[] probs)
    {
        this.N = N;
        this.probs = probs;
         /// Makes sure probabilities add up to one, by calculating a normalization factor
        version(assert)
        {
            T norm = 0;
            foreach(k, p; this.probs)
            {
                norm += p;
            }
            assert(fabs(norm - 1) <= double.min_normal * probs.length * 2);
        }
    }
```
`version(assert)` is always set when asserts are set.

fccoelho: There's another issue: a conservative assumption about the rounding error. See this example:

https://run.dlang.io/is/cVUDcv

The definition of the probs seems fine, but the rounding error is much greater than what is allowed by the assert. The rounding error, `fabs(norm-1)`,  grows much faster than `double.min_normal * probs.length * 2`. The plot below shows the rounding error from probs lengths from 3 to 200: in orange and the function above in green.

image

I think this is a strong reason for not adding this assert.

fccoelho: > Thank you for the work on this MR!

You're welcome, it's been a good opportunity for me to learn a bit of D.

I have replaced `min_normal` by `epsilon` and it's now guaranteed to cover the rounding errors.

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done.

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,105 @@ nothrow @safe version(mir_random_test) unittest
...
+    ///
+    enum isNdRandomVariable = true;
+    ///
+    alias Element = uint;
+
+    ///
+    const(T)[] probs;
+    size_t N;
+    double norm = 0.0;
```
9il: should be removed

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,105 @@ nothrow @safe version(mir_random_test) unittest
...
+        }
+
+    }
+
+    ///
+    pragma(inline, false)
+    void opCall(G)(scope ref G gen, scope size_t[] result)
+        if (isSaturatedRandomEngine!G)
+    {
+        double sum_p = 0.0;
```
9il: T sum_p

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,105 @@ nothrow @safe version(mir_random_test) unittest
...
+MultinomialVariable!(T) multinomialVar(T)(size_t N, return const T[] probs)
+    if (isFloatingPoint!T)
+{
+    return typeof(return)(N, probs);
+}
+
+/// ditto
+alias multinomialVariable = multinomialVar;
+
+nothrow @safe version(mir_random_test) unittest
```
9il: good to make it documented

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,105 @@ nothrow @safe version(mir_random_test) unittest
...
+            norm += p;
+        }
+            assert(fabs(norm - 1) <= T.epsilon * probs.length * 2);
+        }
+
+    }
+
+    ///
+    pragma(inline, false)
+    void opCall(G)(scope ref G gen, scope size_t[] result)
```
9il: as was mentioned before it should be `scope uint[] result` as a user will never use huge arrays for ndvariables.

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,105 @@ nothrow @safe version(mir_random_test) unittest
...
+            }
+
+            sum_p += p;
+            sum_n += result[k];
+        }
+
+
+    }
+    /// ditto
+    void opCall(G)(scope G* gen, scope size_t[] result)
```
9il: ditto
as was mentioned before it should be scope uint[] result as a user will never use huge arrays for ndvariables.

fccoelho: done

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: If I replace this with uint I get this error:
```D
source/mir/random/ndvariable.d(378,7): Error: template mir.random.ndvariable.MultinomialVariable!double.MultinomialVariable.opCall cannot deduce function from argument types !()(MersenneTwisterEngine!(ulong, 64LU, 312LU, 156LU, 31LU, 13043109905998158313LU, 29LU, 6148914691236517205LU, 17LU, 8202884508482404352LU, 37LU, 18444473444759240704LU, 43LU, 6364136223846793005LU), ulong[]), candidates are:
source/mir/random/ndvariable.d(327,10):        mir.random.ndvariable.MultinomialVariable!double.MultinomialVariable.opCall(G)(ref scope G gen, scope uint[] result) if (isSaturatedRandomEngine!G)
source/mir/random/ndvariable.d(353,10):        mir.random.ndvariable.MultinomialVariable!double.MultinomialVariable.opCall(G)(scope G* gen, scope uint[] result) if (isSaturatedRandomEngine!G)
/usr/bin/dmd failed with exit code 1.
```

source/mir/random/ndvariable.d
fccoelho: done

source/mir/random/ndvariable.d
fccoelho: If I remove this I get this error:
```D
source/mir/random/ndvariable.d(378,7): Error: template mir.random.ndvariable.MultinomialVariable!double.MultinomialVariable.opCall cannot deduce function from argument types !()(MersenneTwisterEngine!(ulong, 64LU, 312LU, 156LU, 31LU, 13043109905998158313LU, 29LU, 6148914691236517205LU, 17LU, 8202884508482404352LU, 37LU, 18444473444759240704LU, 43LU, 6364136223846793005LU), ulong[]), candidates are:
source/mir/random/ndvariable.d(327,10):        mir.random.ndvariable.MultinomialVariable!double.MultinomialVariable.opCall(G)(ref scope G gen, scope uint[] result) if (isSaturatedRandomEngine!G)
source/mir/random/ndvariable.d(353,10):        mir.random.ndvariable.MultinomialVariable!double.MultinomialVariable.opCall(G)(scope G* gen, scope uint[] result) if (isSaturatedRandomEngine!G)
/usr/bin/dmd failed with exit code 1.
```

source/mir/random/ndvariable.d
jmh530: I recall this issue with the size_t and uint from before. Would something like below be useful:

```
import std.traits : isUnsigned;

template smaller_unsigned(T)
    if (isUnsigned!T)
{
    import std.meta;
    alias unsigned_types = AliasSeq!(ubyte, ushort, uint, ulong);
    static if (is(T == Largest!(unsigned_types[3], T))) {
        alias smaller_unsigned = unsigned_types[2];
    } else static if (is(T == Largest!(unsigned_types[2], T))) {
        alias smaller_unsigned = unsigned_types[1];
    } else static if (is(T == Largest!(unsigned_types[1], T))) {
        alias smaller_unsigned = unsigned_types[0];
    }
}
```

9il: simple `uint[]` is good for API simplicity. We just need to update the unittests.

source/mir/random/ndvariable.d
9il: the overload with pointer engine should be tested as well

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,106 @@ nothrow @safe version(mir_random_test) unittest
...
+        this.N = N;
+        this.probs = probs;
+         /// Makes sure probabilities add up to one, by calculating a normalization factor
+        version(assert)
+        {
+            T norm = 0;
+            foreach(k, p; this.probs)
+        {
+            norm += p;
+        }
```
9il: +4 spaces

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,106 @@ nothrow @safe version(mir_random_test) unittest
...
+
+    }
+
+    ///
+    pragma(inline, false)
+    void opCall(G)(scope ref G gen, scope size_t[] result)
+        if (isSaturatedRandomEngine!G)
+    {
+        T sum_p = 0.0;
+        uint sum_n = 0;
```
9il: the aggregate type should be `size_t`

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,106 @@ nothrow @safe version(mir_random_test) unittest
...
+/// ditto
+alias multinomialVariable = multinomialVar;
+
+/// Tests if sample returned is of correct size.
+nothrow @safe version(mir_random_test) unittest
+{
+    size_t s = 10000;
+    double[6] p =[1/6., 1/6., 1/6., 1/6., 1/6., 1/6.]; // probs must add up to one
+    auto rv = multinomialVar(s, p);
+    size_t[6] x;
```
9il: uint[6] x;

source/mir/random/ndvariable.d
```diff
@@ -279,6 +280,106 @@ nothrow @safe version(mir_random_test) unittest
...
+    rv(rne,x[]);
```
9il: `rv(rne, x);`

source/mir/random/ndvariable.d
fccoelho: done.

source/mir/random/ndvariable.d
fccoelho: Agreed. Done.

source/mir/random/ndvariable.d
fccoelho: done.

source/mir/random/ndvariable.d
fccoelho: done.

source/mir/random/ndvariable.d
fccoelho: done.

source/mir/random/ndvariable.d
fccoelho: done.

source/mir/random/ndvariable.d
9il: well, then take a look into the function in the error msg and try to get what is going on here
should not be a big problem

fccoelho: These variables need to be declared here because they have their values set in the constructor and are used later in the `opCall`. Other variables in this module use the same pattern. For example `DirichletVariable` does this for its `alpha` parameter.

fccoelho:
Is there anything left to fix here, before it can be merged?

9il:
`norm` member should be removed as we assert in the constructor the norm is about 1.,  opCall shouls just use `1` as well.

PR #120, opened 2020-01-12
fix #119 (Wrong variance for the last dimension)

### LINKED ISSUE #119: Multivariate normal distribution gives wrong variance of second component
I try to generate two dimensional normal distributed point cloud using covariance matrix:
[ 4 0 
  0 4 ]
then I calculate variance for each dimension and for second dimension variance is around 1.0 instead of 4.0

this source code is runnable on run.dlang.org:
```D
/+dub.sdl:
dependency "mir-algorithm" version="~>3.7.8"
dependency "mir-random" version="~>2.2.8"
+/

import std;

void main()
{
    import mir.ndslice.slice: sliced;
    import mir.random;
    import mir.random.ndvariable : multivariateNormalVar;

    // given settings
    enum DataSize = 10_000;
    scope Random* gen = threadLocalPtr!Random;
    auto sigma = [4.0f, -0.0f, -0.0f, 4.0f].sliced(2,2);
    auto rv = multivariateNormalVar!float(sigma);
    
    // generate samples
    float[2][] samples;
    samples.length = DataSize;
    samples.each!((ref v) { rv(gen, v[]); });
    //writeln(samples);
    
    // calculate mean and variance
    const mean_y = samples.map!"a[1]".mean;
    writeln(mean_y);
    
    float var_y = 0;
    samples.each!(v=>var_y += (v[1]-mean_y)^^2);
    var_y /= DataSize-1;
    writeln(var_y);          // <====== variance is expected to be around 4.0, but it always is around 1.0
}
```

drug007:


drug007:
Probably travis-ci faults are not related to PR - I failed to reproduce them using ubuntu 18.04, dmd 2.086 and dmd 2.090.

PR #121, opened 2020-01-16
Upgrade pyd to v0.13.0

Geod24:
Use a newer version which does not depend on class deallocator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,25 @@
-sudo: false
 os:
  - linux
 # Temporarily disable osx builds in Travis CI because of very long wait times
 # - osx
 language: d
 d:
- - ldc
- - dmd
- - ldc-beta
- - dmd-nightly
- - dmd-beta
+ - ldc,dub
+ - dmd,dub
+ - ldc-beta,dub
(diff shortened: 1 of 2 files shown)


Geod24:
Looks like the tests SEGV now :/

Geod24:
Hum it looks like master have been broken for a while. @9il can you take a look ? This blocks a DMD PR.

9il:
No idea why Travis is failing

Geod24:
Unfortunately that won't be enough to unblock the DMD PR.
A new tagged version is needed, which by itself will create other problems if the CI is failing.

So here are the two options I see:
1) Temporarily disable mir-random in `dlang/ci`
2) Fix the CI and make a new point release

The latter would be preferred, but if you don't have the time, I can go with the former. Just let me know what works for you. The PR to DMD is not time-critical, since we're still far away from the next minor release, so there's a bit of time to address the issue.

9il:
tagged with `v2.2.10`, the previous releases failed in the CI anyway

Geod24:
Thanks!

PR #124, opened 2020-02-29
Correct version dependency on mir-core

Geod24:
```
Currently, programs depending on mir-random and mir-algorithm are broken
(e.g. lubeck) because of the version specification.
While it looks very much like a bug in dub, there doesn't seem to be a
tangible reason to have the current version restriction:
if v1.0.0 works, any v1.x.y should work as well.
```

Currently Buildkite builds are broken because of this.


Geod24:
We'll also need a release for the registry to pick it up :(

9il:
`v2.2.11`

PR #129, opened 2020-09-01
-preview=in: Replace 'in ref' with either 'in' or 'const scope ref'

Geod24:
```
Allow the code to compile with '-preview=in'.
In the test code, 'in ref' was replaced with simply 'in',
as the test code isn't much affected by the loss of 'ref'.
However, for the library code, the original meaning was retained.
```

In support of https://github.com/dlang/dmd/pull/11632 (so I can ensure projects compile and the switch is usable).


Geod24:
@9il : Can we get this tagged ?

9il:
done

PR #130, opened 2021-03-21
Add .save property to XoshiroEngine, Xoroshiro128Plus, PermutedCongruentialEngine when stream type is oneseq or none, and applicable instances of PhobosRandom

n8sh:


n8sh:
```d
@property typeof(this) save()() const
{
    typeof(return) copy = void;
    foreach (i, ref fld; this.tupleof)
        copy.tupleof[i] = fld;
    return copy;
}
```

This technique for saving with a disabled postblit worked in `XoshiroEngine` and `Xoroshiro128Plus` but for reasons unclear to me the compiler does not accept it in `mir.random.engine.pcg.PermutedCongruentialEngine`. If it later occurs to me what the problem is I will go back and submit a separate PR to add `.save` to `PermutedCongruentialEngine`.

source/mir/random/engine/pcg.d
```diff
@@ -246,6 +246,17 @@ public:
         void popFront()() { state = bump(state); }
         /// ditto
         void seed()(Uint seed) { this.__ctor(seed); }
+        /// ditto
+        static if (streamType == stream_t.none || streamType == stream_t.oneseq)
+        @property typeof(this) save()() const
+        // Repeat the condition so it shows up in DDoc.
+        if (streamType == stream_t.none || streamType == stream_t.oneseq)
```
9il, line 253: the condition duplicates `static if` above

n8sh: I did that because the auto-generated DDoc will include the function without any mention of `static if`. I'll remove the redundant constraint and instead add a note in the documentation about when `save` is available.

9il:
Feel free to make a patch release if required.

PR #131, opened 2021-03-21
Add .save property to PermutedCongruentialEngine when stream type is stream_t.specific

n8sh:
Followup to #130.

### DIFF
--- a/source/mir/random/engine/pcg.d
+++ b/source/mir/random/engine/pcg.d
@@ -226,11 +226,12 @@ public:
         /++
         Compatibility with $(LINK2 https://dlang.org/phobos/std_random.html#.isUniformRNG,
         Phobos library methods). Presents this RNG as an InputRange.
-        Only available if `output_previous == true`. `save` is available when
-        `streamType` is `stream_t.none` or `stream_t.oneseq`.
+        Only available if `output_previous == true`.
 
         The reason that this is enabled when `output_previous == true` is because
         `front` can be implemented without additional cost.
+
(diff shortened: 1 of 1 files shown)


source/mir/random/engine/pcg.d
```diff
@@ -434,9 +429,7 @@ mixin template specific_stream(Uint)
     ///
     Uint inc_ = default_increment!Uint;
     ///
-    @property Uint increment()() { return inc_; }
-    ///
-    @property void increment()(Uint u) { inc_ = u;}
+    alias increment = inc_;
```
n8sh, line 432: For some reason `@property Uint increment()() { return inc_; }` was causing `save` not to work. Changing it to a non-template property function was sufficient for `save` to work but I went a step further and replaced the property with an alias.

n8sh: Related bug report https://issues.dlang.org/show_bug.cgi?id=21744

source/mir/random/engine/pcg.d
```diff
@@ -248,8 +249,8 @@ public:
         /// ditto
         void seed()(Uint seed) { this.__ctor(seed); }
         /// ditto
-        static if (streamType == stream_t.none || streamType == stream_t.oneseq)
```
9il, line 251: This needs to be preserved for `hasMember` traits, otherwise, it will return true while it should return false.

n8sh: Fixed.

